PR #2, opened 2025-05-11
Revert "Bump com.github.tomakehurst:wiremock-jre8 from 2.35.0 to 2.35.1 in /core in the maven group across 1 directory"

cornellgit:
Reverts google/adk-java#1


cornellgit:
discard

PR #155, opened 2025-05-31
chore: Introduce a parent POM

vorburger:


core/pom.xml
```diff
@@ -14,86 +14,35 @@
...
 -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
 
+  <parent>
+    <groupId>com.google.adk</groupId>
+    <artifactId>google-adk-parent</artifactId>
+    <version>0.1.1-SNAPSHOT</version><!-- {x-version-update:google-adk:current} -->
```
shukladivyansh, line 25: Can we keep the version same as before, 0.1.0

shukladivyansh (commented):
I remember @Poggecci was running into some issue previously when trying to unify the poms. Maybe we would want to check for those.

core/pom.xml
Poggecci: Issues were just that though building the parent pom would WAI, then the children poms wouldn't be able to build on their own. 

For @vorburger:
You said that it's possible to still deploy the submodules independently, is this a trivial thing already possible with the current code proposed in this PR, or does this require additional setup?

vorburger: > Can we keep the version same as before, 0.1.0

We could, but we really shouldn't... 😆 but for other reasons, and technically it has nothing to do with having a parent POM or not. Tell you what, I'll raise a separate PR with JUST containing this, including details about WHY we need to do this. Then we can discuss this specific aspect there.

> Issues were just that though building the parent pom would WAI, then the children poms wouldn't be able to build on their own. You said that it's possible to still deploy the submodules independently, is this a trivial thing already possible with the current code proposed in this PR, or does this require additional setup?

With this (as-is), you could already do either `./mvnn test -f core` or `cd core && ../mvnw test`, and then it would still only build the `core` module - and not also the `dev` module, which it [now, new] would automagically do together if you did `./mvnw test` in the root directory. Is this what you meant?

Or did you mean if it's possible to do `./mvnw deploy` (!) of `core` & `dev` separately? That may not work - depending on how "smart" Maven is to detect a `SNAPSHOT` parent POM... I have never tried doing this. If Maven is "smart" and refuses it (which it should!), then in theory you most likely could first release the parent POM, and then adjust the version number in a module, and deploy that. But this approach is a Very Bad Idea - IMHO. It's confusing like hell to end-users. As in: "Wait so these guys have 1 Git repo, and so now which version of what JAR is from which branch... OMG, this is crazy!". Similarly for maintainers. This isn't the Google monorepo, and Bazel. This is a tiny git repo, with a bunch of Java classes, which are packaged up into 2 JARs, which are pushed to the binary Maven repository commonly known as Maven Central. Release such 2 JARs, or even 3 or even 5, from the same Git repo, with separate version numbers, might sound fun and cool- but it's a completely pointless overhead. Let's just regularly release the (few, handful of) modules of a single git repo under a common version, together? There really is no very good reason, that I know of, to do this any differently. Most sensible Java Maven projects do (should) this like that. If you have specific reasons why you think that may be a Very Bad Idea, I'd love to be given an opportunity to provide counter arguments for any concerns related to proceeding like this.

@glaforge FYI

vorburger:
> Approved contingent on fixing the merge conflicts.

@Poggecci conflicts now resolved; merge?

PR #178, opened 2025-06-08
fix: Remove wrong auto-value dependency (fixes #172)

### LINKED ISSUE #172: {P2} Annotation processing is enabled because ...
**Describe the bug**

```
$ git clone https://github.com/enola-dev/LearningADK.git
$ cd quickstart
$ ../mvnw clean compile

(...)

[INFO] --- compiler:3.14.0:compile (default-compile) @ adk-agents ---
[INFO] Recompiling the module because of changed source code.
[INFO] Compiling 1 source file with javac [debug target 21] to target/classes
[INFO] Annotation processing is enabled because one or more processors were found
  on the class path. A future release of javac may disable annotation processing
  unless at least one processor is specified by name (-processor), or a search
  path is specified (--processor-path, --processor-module-path), or annotation
  processing is enabled explicitly (-proc:only, -proc:full).
  Use -Xlint:-options to suppress this message.
  Use -proc:none to disable annotation processing.
```

This doesn't really cause any harm, but it's... still at least a bit "surprising", no?

I wonder what JAR pulls along an (unused, clearly) APT, and how we might exclude it.

@glaforge FYI

vorburger:
Fixes #172.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -134,12 +134,6 @@
     </dependency>
     <dependency>
       <groupId>com.google.auto.value</groupId>
-      <artifactId>auto-value</artifactId>
-      <version>${auto-value.version}</version>
-      <optional>true</optional>
-    </dependency>
-    <dependency>
-      <groupId>com.google.auto.value</groupId>
       <artifactId>auto-value-annotations</artifactId>
       <version>${auto-value.version}</version>
       <scope>provided</scope>


shukladivyansh:
hi michael!

can you add some explanation on why the removal? i see we still have have autovalue dependency on `auto-value-annotations` and on `auto-value` for annotation processing. Are these the only 2 needed?

vorburger:
> can you add some explanation on why the removal?

@shukladivyansh the "rationale" for this change is in #172, this fixes that. Sorry, but can I double check if you read that? Tx!

> i see we still have have autovalue dependency on auto-value-annotations and on auto-value for annotation processing. Are these the only 2 needed?

Yep, those need to stay! (Otherwise we would break that `@AutoValue` stuff.)

But this (only) needs to go. It won't break the `@AutoValue` stuff.

PS: https://github.com/google/auto/blob/main/value/userguide/index.md has some background, in case you are truly passionate about wanting to fully understand this? 🤣 Note how that doc also recommends (correctly) to use `auto-value` for `annotationProcessorPaths` but NOT for `dependency` (which we are doing, which is wrong, which this removes).

Please let me know if I can provide any further background on this? ("I know what I'm doing." (TM)(R)(C) ... 🔢  but always very happy to explain and share!)

vorburger:
I'm closing this PR as a duplicate of the (bitwise identical) #196.

Issue #168, opened 2025-06-06
Verify Code Format as well as License Header presence during CI build


vorburger:
I'm raising this issue prompted by https://github.com/google/adk-java/pull/102/files#r2132521541:

**Is your feature request related to a problem? Please describe.**

IMHO it would make the life of committers (maintainers) easier if this project would verify Code Format as well as missing License Header presence during CI build (and thus cause red build failures).

IMHO it would also make the life of contributors (PR raising community members) easier if this project had a very easy way to auto-format code according to the standards that this projects expects its code to be formatted with.

**Describe the solution you'd like**

The following is for discussion, but if it was up to me, I would go about this simply by adding a suitable pre-commit hook configuration, for both the Google Java Code Formatters, and the license headers checker/adder. (I have used and done this in a project of mine, and ready to copy/paste this here.) 

**Describe alternatives you've considered**
Waste time of both committers and contributors by leaving "nit pick" comments about such things on PRs.

Using pre-commit would avoid requiring Python to be installed to run the pre-commit, but IMHO nowadays that shouldn't really be an issue. (And one could still locally build the project without it; just without the pre-commit.) 

An alternative to avoid ^^^ would be to integrate the Google Java Code Formatters using (one of) its Maven Plugins. But I don't know (haven't looked into) a Maven license header checker, and as far as I know that's harder to easily cleanly integrate nicely as an automagic pre-commit.

**Additional context**

I'm happy to help by proposing a PR for this, if the project's committers concur that this would be useful?

@glaforge @Poggecci @cornellgit do you have any thought about / input on this?

glaforge:
We could set up checkstyle since committers will be building with Maven?

vorburger:
Checkstyle is obviously great, and I've happily used it for 25 years! But I suspect that contributions to the code in this repository will very likely be expected to be formatted with Google's Java Format (GJF).  And https://github.com/google/google-java-format actually really is pretty awesome! (I've even made some, very very minor, contributions to it myself!) And keeping a Checkstyle config manually aligned with the format that  Google's Java Format expects is going to be a huge unmaintainable maintenance nightmare, IMHO. So personally I would recommand that we simply adopt (hear, enforce) GJF here in this project. Thoughts?

PS: Just to avoid confusion, the "Java/Building with Maven" vs. https://pre-commit.com (Python) dimension is tehcnically completely orthogonal to GFM or not (because we could integrate GFM validation and autoformat through various mechanisms).

glaforge:
[...]

PR #183, opened 2025-06-10
ci: add license validation step for core and dev modules

xinhuagu:
### Description

adds automated license validation steps in CI 

( Hello, If the java.header and license-checks files are automatically used elsewhere or have other uses and it shouldn't be validated in ci, please close this PR. Thanks )

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -15,6 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        module: [core, dev]
         java-version: ['17', '21']
 
     steps:
@@ -35,6 +36,10 @@ jobs:
           restore-keys: | # Fallback keys if exact match is not found
             ${{ runner.os }}-maven-${{ matrix.java-version }}-
             ${{ runner.os }}-maven-
+
+      - name: License validation for ${{ matrix.module }}
+        working-directory: ./${{ matrix.module }}
+        run: mvn checkstyle:check -Dcheckstyle.config.location=../license-checks.xml -Dcheckstyle.header.file=../java.header
 
       - name: Compile and install core module with Java ${{ matrix.java-version }}
         working-directory: ./core
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -39,7 +39,7 @@ jobs:
 
       - name: License validation for ${{ matrix.module }}
         working-directory: ./${{ matrix.module }}
-        run: mvn checkstyle:check -Dcheckstyle.config.location=../license-checks.xml -Dcheckstyle.header.file=../java.header
+        run: mvn checkstyle:check -Dcheckstyle.config.location=../license-checks.xml
 
       - name: Compile and install core module with Java ${{ matrix.java-version }}
         working-directory: ./core
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -416,6 +416,7 @@
         <artifactId>maven-checkstyle-plugin</artifactId>
         <version>${checkstyle-maven-plugin.version}</version>
         <configuration>
+          <propertyExpansion>checkstyle.header.file=../java.header</propertyExpansion>
           <includeTestSourceDirectory>false</includeTestSourceDirectory>
           <failOnViolation>true</failOnViolation>
           <violationSeverity>warning</violationSeverity>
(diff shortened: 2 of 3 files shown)


vorburger:
@xinhuagu this could be useful for #168, but are you sure that this will actually work?

This PR basically adds `mvn checkstyle:check -Dcheckstyle.config.location=../license-checks.xml -Dcheckstyle.header.file=../java.header`, but... there (currently) is [no Checkstyle configured in this project](https://github.com/search?q=repo%3Agoogle%2Fadk-java%20checkstyle&type=code) - agreed? So... this will actually fail to build (I think; haven't tested it yet).

xinhuagu:
> This PR basically adds `mvn checkstyle:check -Dcheckstyle.config.location=../license-checks.xml -Dcheckstyle.header.file=../java.header`, but... there (currently) is [no Checkstyle configured in this project](https://github.com/search?q=repo%3Agoogle%2Fadk-java%20checkstyle&type=code) - agreed? So... this will actually fail to build (I think; haven't tested it yet).

@vorburger  It will build.  The "check" goal was directly invoked with the checkstyle.header.file parameter, even without explicit configuration in the pom.xml. It still works, however, it is the minimal change, not  the optimal one.
I used checkstyle:check, which pulls the default version from Maven Central and there is an undefined property in  license-checks.xml. To ensure clarity and consistency, it's indeed better to define and config the Checkstyle plugin explicitly in the pom.xml. 

I updated the PR accordingly,thanks!

.github/workflows/validation.yml
```diff
@@ -36,6 +37,10 @@ jobs:
             ${{ runner.os }}-maven-${{ matrix.java-version }}-
             ${{ runner.os }}-maven-
 
+      - name: License validation for ${{ matrix.module }}
+        working-directory: ./${{ matrix.module }}
+        run: mvn checkstyle:check -Dcheckstyle.config.location=../license-checks.xml
```
vorburger, line 42: Let's not do this like this (i.e. running Maven twice using a Matrix), but instead wait for https://github.com/google/adk-java/pull/155 to get merged (hopefully in the next few days already), and then simply use standard Maven? (Once there is a parent POM, Checkstyle can run on all modules.)

core/pom.xml
```diff
@@ -63,6 +63,7 @@
     <java-websocket.version>1.6.0</java-websocket.version>
     <jackson.version>2.19.0</jackson.version>
     <okhttp.version>4.12.0</okhttp.version>
+    <checkstyle-maven-plugin.version>3.6.0</checkstyle-maven-plugin.version>
```
vorburger, line 66: Let's wait for https://github.com/google/adk-java/pull/155 and then add this to the (new) root `pom.xml` instead of duplicating it in both the `core/pom.xml` as well as the `dev/pom.xml` ?

core/pom.xml
```diff
@@ -410,6 +411,17 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-checkstyle-plugin</artifactId>
+        <version>${checkstyle-maven-plugin.version}</version>
+        <configuration>
+          <propertyExpansion>checkstyle.header.file=../java.header</propertyExpansion>
```
vorburger, line 419: I'm curious about something that I have not yet understood: Will this ONLY check the `java.header` for the license? That would be a great! Or will this ALSO validate "code format"? That would a problem. Because the code in this repo will never (exactly) match Checkstyle code conventions. (Instead, I will add another code format checker later, see #168.) Could you double check to clarify? Tx!

vorburger (changes requested):

.github/workflows/validation.yml
xinhuagu: Sure, using a parent POM is definitely the better solution

core/pom.xml
xinhuagu: Agree, once the root POM is there, there's no need to repeat the version in multiple places.

core/pom.xml
xinhuagu: this checkstyle.header.file=../java.header will also valide other default rules. That would be a problem , true. Thx for pointing it out. 
 
 looking forward to #168 , i close this pr now :) thx

Issue #184, opened 2025-06-10
Potential version mismatch between Version class and pom.xml


xinhuagu:
### Description
If I understand correctly, the com.google.adk.Version class includes a version number intended to align with the project version defined in pom.xml. Currently, there appears to be a discrepancy between these two version numbers.
```java
package com.google.adk;

/** Holding class for the version of the Java ADK. */
public final class Version {
  public static final String JAVA_ADK_VERSION = "0.1.0";

  private Version() {}
}
```
in current pom.xml
```xml
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.google.adk</groupId>
  <artifactId>google-adk</artifactId>
  <version>0.1.1-SNAPSHOT</version><!-- {x-version-update:google-adk:current} -->
  <packaging>jar</packaging>
  <name>Agent Development Kit</name>
```
### Possible solution
automatically synchronize the Version class with the pom.xml project version during the build phase

vorburger:
> If I understand correctly, the com.google.adk.Version class includes a version number intended to align with the project version defined in pom.xml. Currently, there appears to be a discrepancy between these two version numbers.

FYI this is new, since #170.

> automatically synchronize the Version class with the pom.xml project version during the build phase

That SGTM, in principle. I'll engage on #187 about it.

PR #185, opened 2025-06-10
feat: Use templating-maven-plugin to sync project version in Version.…

xinhuagu:
#184 

uses Maven Java Template Plugin to auto-sync Version class with pom.xml.


xinhuagu:
filtering the standard src/main/java folder is not supported in Templating Maven Plugin

https://www.mojohaus.org/templating-maven-plugin/examples/source-filtering.html

PR #187, opened 2025-06-11
Load version from properties

xinhuagu:
#184 

Add a version.properties file with Maven filtering so that the Version class can dynamically load the project version


vorburger:
I'm generally in favor of (something like) this. Next steps:

1. A maintainer (which I am not, yet) needs to approve building this PR, so that we can see if it's green.

2. We then may have to still verify something with others users of this project to ensure that it doesn't break anything.

shukladivyansh:
Lgtm. Thanks for the change @xinhuagu

vorburger:
LGTM as-is!

PR #203, opened 2025-06-18
(chore) Update LangChain4j to the newly released 1.1.0 BOM

glaforge:
LangChain4j has just cut a new release.
This PR updates the version of the LC4j BOM.

PR #206, opened 2025-06-19
clean: Simplify contrib/langchain4j/pom.xml

vorburger:
As discussed & promised [here](https://github.com/google/adk-java/pull/102#discussion_r2154845230).

Note that, despite removing all of this, Maven will still do exactly the same as before - simply because all of these settings are (now) inherited from the (new) parent POM.

@glaforge Plz R and if LGTY, then click Approve?


vorburger:
This PR now includes the commit from #203 (just to simplify merging, because that would have created a conflict).

PR #208, opened 2025-06-20
chore: Reformat contrib/langchain4j (re. #168)

vorburger:
Re. #168.

This is a purely mechanical reformat. It does not change any semantics, and cannot break anything.

@glaforge Plz R - LGTY?


glaforge (approved):
LGTM

PR #272, opened 2025-07-16
feat: pass timeout config while init mcp client

wenhaozhao:
allow to pass timout config while create mcp-client


vorburger:
> allow to pass timout config while create mcp-client

@wenhaozhao can you elaborate on how this would be used? 

> if (connectionParams instanceof SseServerParameters sseServerParams)

How would one configure this? It might be useful to document this.

PS: Looks like this could be a solution for #285 from @rsaulo.

wenhaozhao:
> > allow to pass timout config while create mcp-client
> 
> @wenhaozhao can you elaborate on how this would be used?
> 
> > if (connectionParams instanceof SseServerParameters sseServerParams)
> 
> How would one configure this? It might be useful to document this.
> 
> PS: Looks like this could be a solution for #285 from @rsaulo.

        McpSessionManager mcpManager = new McpSessionManager(SseServerParameters.builder()
                .url("http://127.0.0.1:28080/sse")
                .timeout(Duration.ofSeconds(120L)) //set init timeout
                .sseReadTimeout(Duration.ofSeconds(600L)) // set read timout
                .build());
        var mcpClient = mcpManager.createSession();
        var mcpTools = mcpClient.client().listTools().tools().stream()
                .map(tool -> new McpTool(tool, mcpClient, mcpManager))
                .toList();

wenhaozhao:
> > allow to pass timout config while create mcp-client
> 
> @wenhaozhao can you elaborate on how this would be used?
> 
> > if (connectionParams instanceof SseServerParameters sseServerParams)
> 
> How would one configure this? It might be useful to document this.
> 
> PS: Looks like this could be a solution for #285 from @rsaulo.

McpSyncClient is a wrapper around McpAsyncClient that provides a blocking API via block(). When sending requests, the timeout parameter becomes effective within the reactor-chain.

image

so, you can config timeout like this:

        McpSessionManager mcpManager = new McpSessionManager(SseServerParameters.builder()
                .url("http://127.0.0.1:28080/sse")
                .timeout(Duration.ofSeconds(120L)) //set init timeout
                .sseReadTimeout(Duration.ofSeconds(600L)) // set read timout
                .build());
        var mcpClient = mcpManager.createSession();
        var mcpTools = mcpClient.client().listTools().tools().stream()
                .map(tool -> new McpTool(tool, mcpClient, mcpManager))
                .toList();

wenhaozhao:
> > allow to pass timout config while create mcp-client
> 
> @wenhaozhao can you elaborate on how this would be used?
> 
> > if (connectionParams instanceof SseServerParameters sseServerParams)
> 
> How would one configure this? It might be useful to document this.
> 
> PS: Looks like this could be a solution for #285 from @rsaulo.

and there a another pr https://github.com/google/adk-java/pull/273, also include timeout config

vorburger:
> so, you can config timeout like this:

Awesome! We should probably document this, somehow somewhere? Would you be willing to raise a PR to add a short section e.g. on https://github.com/google/adk-docs/blob/main/docs/get-started/streaming/quickstart-streaming-java.md for https://google.github.io/adk-docs/get-started/streaming/quickstart-streaming-java/ about this?

vorburger:
> > so, you can config timeout like this:
> 
> Awesome! We should probably document this (...) https://google.github.io/adk-docs/get-started/streaming/quickstart-streaming-java/ about this?

Probably better on https://google.github.io/adk-docs/tools/mcp-tools, actually?

wenhaozhao:
> > > so, you can config timeout like this:
> > 
> > 
> > Awesome! We should probably document this (...) https://google.github.io/adk-docs/get-started/streaming/quickstart-streaming-java/ about this?
> 
> Probably better on https://google.github.io/adk-docs/tools/mcp-tools, actually?


 https://google.github.io/adk-docs/tools/mcp-tools maybe better, but it's written in python.
Can you add java sections for me?

core/src/main/java/com/google/adk/tools/mcp/McpSessionManager.java
```diff
@@ -51,16 +52,21 @@ public static McpSyncClient initializeSession(Object connectionParams) {
...
     McpClientTransport transport = transportBuilder.build(connectionParams);
-
+    if (connectionParams instanceof SseServerParameters sseServerParams) {
+      initializationTimeout = sseServerParams.timeout();
+      requestTimeout = sseServerParams.sseReadTimeout();
+    }
     McpSyncClient client =
         McpClient.sync(transport)
-            .requestTimeout(Duration.ofSeconds(10))
+            .initializationTimeout(Optional.ofNullable(initializationTimeout).orElse(Duration.ofSeconds(10)))
```
vorburger: This line is causing a build failure because it's not code formatted correctly, see https://github.com/google/adk-java/actions/runs/16477018497/job/46581287119?pr=272 ... it should auto-format if you rebase and re-run the Maven build locally - can you do that, and then force push it? (You can just overwrite the merge commit that I created.)

vorburger (changes requested):
Could you address the build failure due to the code format? (It's something we just added.)

core/src/main/java/com/google/adk/tools/mcp/McpSessionManager.java
wenhaozhao: I've fix it, and force push it again, please review again

wenhaozhao:
> Could you address the build failure due to the code format? (It's something we just added.)

fine, i've resloved

core/src/main/java/com/google/adk/tools/mcp/McpSessionManager.java
vorburger: 👍🏽

vorburger:
> https://google.github.io/adk-docs/tools/mcp-tools maybe better, but it's written in python. Can you add java sections for me?

Watch new issue #297 re. the missing doc (in general about MCP, not about this timeout); maybe once that's resolved, you could add this.

rsaulo:
Just tested with main branch and it is working.

For me is working and it is OK.

Just a comment, now I see that in McpSessionManager, the class uses the sseReadTimeout parameter from SseServerParameters to set the requestTimeout, BUT, this parameter has a default of 5 MINUTES on the SseServerParameters builder, what is turning this parameter as 5 minutes forever. I understand that we need a specific parameter on SseServerParameters like "mcpClientRequestTimeout" to use here as sseReadTimeout is not the same as requestTimeout. But for my use case it is working!! Thanks.

@vorburger , thanks for the quick attention on this matter.

PR #273, opened 2025-07-16
feature: support async mcp tool client

wenhaozhao:
support  async-mcp-client


core/src/main/java/com/google/adk/tools/mcp/McpAsyncTool.java
```diff
@@ -0,0 +1,149 @@
...
+   * @param mcpSessionManager The MCP session manager to use to create new sessions.
+   * @throws IllegalArgumentException If mcpTool or mcpSession are null.
+   */
+  public McpAsyncTool(
+      Tool mcpTool, Single<McpAsyncClient> mcpSession, McpSessionManager mcpSessionManager) {
+    this(mcpTool, mcpSession, mcpSessionManager, JsonBaseModel.getMapper());
+  }
+
+  /**
+   * Creates a new McpTool with the default ObjectMapper.
```
Poggecci: not the default object mapper here as you are passing one in explicitly

core/src/main/java/com/google/adk/tools/mcp/McpAsyncTool.java
```diff
@@ -0,0 +1,149 @@
...
+   *
+   * @param mcpTool The MCP tool to wrap.
+   * @param mcpSession The MCP session to use to call the tool.
+   * @param mcpSessionManager The MCP session manager to use to create new sessions.
+   * @param objectMapper The ObjectMapper to use to convert JSON schemas.
+   * @throws IllegalArgumentException If mcpTool or mcpSession are null.
+   */
+  public McpAsyncTool(
+      Tool mcpTool,
+      Single<McpAsyncClient> mcpSession,
```
Poggecci: why have this be a Single?

Poggecci (changes requested):
Thank you a lot for this contribution!
Would you mind updating the description briefly detailing why the Async client was necessary for your use case? Would love to better understand and see if we should update our docs to default to the Async over the sync client

wenhaozhao:
> Thank you a lot for this contribution! Would you mind updating the description briefly detailing why the Async client was necessary for your use case? Would love to better understand and see if we should update our docs to default to the Async over the sync client

Thanks for you relay.


I am building an Agent service based on Java, providing RESTful APIs through Spring Boot Webflux. ADK-Java meets my needs. Meanwhile, it is developed based on RxJava. Although it and Reactor in Webflux are two different reactive frameworks, it's not a problem. Asynchronous and reactive technology stacks are exactly what I want. But in the process of me gradually delving into using ADK, I found that in the link based on the asynchronous technology stack:

1 McpTool is actually an implementation based on McpSyncClient.
2 And mcp-java-sdk is an asynchronous implementation based on Reactor, providing asynchronous APIs through McpAsyncClient. McpSyncClient converts asynchronous APIs to synchronous APIs by means of Mono.block(). 
Thus, the call link of Async -> Sync -> Async is formed for McpTool in ADK. 
However, the key point is that using blocking operations in an asynchronous context may lead to performance problems and waste of thread resources. 
So, I decided to provide an asynchronous implementation of McpTool to better match the asynchronous technology stack architecture of ADK.

core/src/main/java/com/google/adk/tools/mcp/McpAsyncTool.java
wenhaozhao: Because the initialization process takes place in McpSessionManager.initializeAsyncSession(...), and when using McpAsyncClient without calling block(), the initialization result is returned within the asynchronous chain. Therefore, McpSessionManager.initializeAsyncSession needs to return an asynchronous-wrapped Single<McpAsyncClient>. 
Also, there is a reinitializeSession() process in McpAsyncTool that needs to be supported. However, in the latest commit, the initialization process has been adjusted to McpAsyncTool.reinitializeSession() and integrated into the relevant chain, so the wrapping with Single<...> is removed in the latest commit.

core/src/main/java/com/google/adk/tools/mcp/McpAsyncTool.java
wenhaozhao: i've fix this comment

wenhaozhao:
> Thank you a lot for this contribution! Would you mind updating the description briefly detailing why the Async client was necessary for your use case? Would love to better understand and see if we should update our docs to default to the Async over the sync client



> > Thank you a lot for this contribution! Would you mind updating the description briefly detailing why the Async client was necessary for your use case? Would love to better understand and see if we should update our docs to default to the Async over the sync client
> 
> Thanks for you relay.
> 
> I am building an Agent service based on Java, providing RESTful APIs through Spring Boot Webflux. ADK-Java meets my needs. Meanwhile, it is developed based on RxJava. Although it and Reactor in Webflux are two different reactive frameworks, it's not a problem. Asynchronous and reactive technology stacks are exactly what I want. But in the process of me gradually delving into using ADK, I found that in the link based on the asynchronous technology stack:
> 
> 1 McpTool is actually an implementation based on McpSyncClient. 2 And mcp-java-sdk is an asynchronous implementation based on Reactor, providing asynchronous APIs through McpAsyncClient. McpSyncClient converts asynchronous APIs to synchronous APIs by means of Mono.block(). Thus, the call link of Async -> Sync -> Async is formed for McpTool in ADK. However, the key point is that using blocking operations in an asynchronous context may lead to performance problems and waste of thread resources. So, I decided to provide an asynchronous implementation of McpTool to better match the asynchronous technology stack architecture of ADK.

So, I think it's a better choice to use asynchronous clients in the asynchronous link.

Poggecci (approved):

andrey-star:
Can we also either update [core/src/main/java/com/google/adk/tools/mcp/McpToolset.java](https://github.com/google/adk-java/blob/main/core/src/main/java/com/google/adk/tools/mcp/McpToolset.java) to allow choosing, or create an `McpAsyncToolset`?

wenhaozhao:
In my opinion, it might be better to create an McpAsyncToolset first, and then gradually and smoothly migrate the McpToolset to McpAsyncToolset.

wenhaozhao:
I'll add McpAsyncToolset later

wenhaozhao:
> Can we also either update [core/src/main/java/com/google/adk/tools/mcp/McpToolset.java](https://github.com/google/adk-java/blob/main/core/src/main/java/com/google/adk/tools/mcp/McpToolset.java) to allow choosing, or create an `McpAsyncToolset`?

The pr for McpAsyncToolset: https://github.com/google/adk-java/pull/306

PR #279, opened 2025-07-17
chore: fix formatting

kaikreuzer:


vorburger (changes requested):
Let's not merge this just yet - I'd like to better understand why this wasn't automatically done already; see https://github.com/google/adk-java/issues/168#issuecomment-3089427206; I'll try to have a closer look into this some time.

vorburger:
I've had a quick first closer look at this, so:

Reformatting `LangChain4jTest` is CORRECT, but in the future hopefully we shouldn't need manual PRs like this, but our CI should  fail for ill-formatted Java code? I'll see if that's easy enough to fix... i.e. running Maven should IMHO be made to FAIL when it fixes the formatting (at least on CI).

The change currently proposed to re-order those 2 imports in the `Runner.java` appears wrong, to me. Running `./mvnw clean test` (on 5a7ab202ae26047796d8ff8a5d6bffcebbb6e180) does NOT cause that (contrary to the change in `LangChain4jTest`) for me... @kaikreuzer might that be a change you inadvertently included by mistake in this PR?

kaikreuzer:
> @kaikreuzer might that be a change you inadvertently included by mistake in this PR?

No, definitely not.
I did a clean checkout and ran `mvn clean test` on it. This PR includes exactly the changes that were done when running the Maven build.

vorburger:
> No, definitely not. I did a clean checkout and ran `mvn clean test` on it. This PR includes exactly the changes that were done when running the Maven build.

Ah, that's just because with & without https://github.com/google/adk-java/commit/5a7ab202ae26047796d8ff8a5d6bffcebbb6e180...

vorburger:
> Reformatting LangChain4jTest is CORRECT, but in the future hopefully we shouldn't need manual PRs like this, but our CI should fail for ill-formatted Java code? I'll see if that's easy enough to fix... i.e. running Maven should IMHO be made to FAIL when it fixes the formatting (at least on CI).

see https://github.com/google/adk-java/pull/282 for a proposal how to address that problem.

PR #290, opened 2025-07-20
feat: Add CORS configuration for local ADK-Web angular

xinhuagu:
This PR enables CORS configuration in the ADK Java dev ui server to allow requests from localhost:2400, which is the default development port for the adk-web Angular UI. (https://github.com/google/adk-web)

Unlike adk-python, which provides CLI commands such as adk api_server and adk web, the Java implementation requires starting the server manually by maven on port 8000, which is expected by the adk-web. 

Since it is not straightforward to pass CORS configuration via the "mvn compile exec:java" command line, this PR adds a configuration-based solution to enable CORS support for localhost:2400


dev/src/main/java/com/google/adk/web/config/LocalAdkWebCorsConfig.java
```diff
@@ -0,0 +1,22 @@
...
+ * Configuration class for connecting the local ADK-Web.
+ */
+@Configuration
+public class LocalAdkWebCorsConfig implements WebMvcConfigurer {
+
+  @Override
+  public void addCorsMappings(CorsRegistry registry) {
+    registry
+        .addMapping("/**")
+        .allowedOrigins("http://localhost:4200")
```
vorburger: @xinhuagu I'm personally not a huge fan of seeing this completely hard-coded to `4200`... might there be a way to make this default to 4200, but use some fancy Spring configuration thing to keep it easily customizable?

vorburger (changes requested):

dev/src/main/java/com/google/adk/web/config/LocalAdkWebCorsConfig.java
xinhuagu: > @xinhuagu I'm personally not a huge fan of seeing this completely hard-coded to `4200`... might there be a way to make this default to 4200, but use some fancy Spring configuration thing to keep it easily customizable?

@vorburger Yes, that's a great idea! I've updated it to use application.yml for configuration, which should make it much more flexible. Thanks so much for the helpful tip! Please let me know if there's anything else I should adjust.

xinhuagu: Now, without a cors setting in application.yml, the following mvn command works as well:

```
mvn compile exec:java "-Dexec.args=--server.port=8000 \
     --adk.web.cors.origins=http://localhost:4200 \
     --adk.agents.source-dir=src/ \
     --logging.level.com.google.adk.dev=DEBUG"
```

dev/src/main/resources/application.yml
```diff
@@ -0,0 +1,20 @@
...
+        - "DELETE"
+        - "OPTIONS"
+      headers:
+        - "*"
+      allow-credentials: true
+      max-age: 3600
+
+logging:
+  level:
+    com.google.adk.dev: DEBUG
```
vorburger: This will enable `DEBUG` level logging for all of `adk.dev` by default for all users... could be noisy?

Is this intentional, or perhaps just a left over from your local testing?

How about removing this, as its technically unrelated to CORS?

dev/src/main/resources/application.yml
```diff
@@ -0,0 +1,20 @@
+adk:
+  web:
+    cors:
+      mapping: "/**"
+      origins:
+        - "http://localhost:4200" # Allow local ADK-Web Angular development server
```
vorburger: Wondering if really want this here... as this is in `src/main/resources`, doesn't this "effectively" make it _just like_ it was hard-coded - as it was in the original iteration when it was in Java code? Or maybe not quite, as its configurable now... but by default its still on. Do we want that? Are there any security implications with this? It's also not consistent with the explanation given in the _Conversation_  about using `--adk.web.cors.origins=http://localhost:4200`... if this were here, then that wouldn't be needed, right? My personal preference would actually be to allow enable users to configure CORS - but not having this be the default - wouldn't that be... "safer"?

Is this intentional, or perhaps just a left over from your local testing?

dev/src/main/java/com/google/adk/web/config/LocalAdkWebCorsConfig.java
vorburger: > Now, without a cors setting in application.yml, the following mvn command works as well:

I like this! But I'm slightly confused, because in the current version there IS a CORS setting in a default `application.yaml`... is that intentional, or was it just something left over which you meant to remove? Personally I would prefer NOT having it configured and enabled by default, and use this approach - which we should document? Would you be willing to raise a PR to add a short section e.g. on https://github.com/google/adk-docs/blob/main/docs/get-started/streaming/quickstart-streaming-java.md for https://google.github.io/adk-docs/get-started/streaming/quickstart-streaming-java/ about this?

xinhuagu: Sorry for the confusion, I agree, better without pre-configuration. I’ve removed the entire application.xml 

> Would you be willing to raise a PR to add a short section e.g. on https://github.com/google/adk-docs/blob/main/docs/get-started/streaming/quickstart-streaming-java.md for https://google.github.io/adk-docs/get-started/streaming/quickstart-streaming-java/ about this?

I’ve been thinking about how to include this cors setting in the quickstart-streaming-java.md doc. Since cors is only needed when connecting with an external adk-web ui server, and the current quickstarts doc already quite complete, unless introduce a section specifically about such external ui integration...

Perhaps an idea that might be helpful: if **adk-java** had its own cli tool similar to the _adk_ command in **adk-python** cli then it could be something like: ( same as _adk api_server_ command)
```
adkj api_server --allow_origins=http://localhost:4200
```
In this case, the _adk.web.cors.origins_ property would correspond to the _--allow_origins_ flag, making it easier to understand and configure. Users wouldn’t need to remember Maven commands or internal property names , just a simple cli command to start the web with the right settings.

A simple and possible cli implementation maybe like : for example use picocli to execute maven goals programmatically using the maven Invoker api within picocli command and install it via jbang ....

dev/src/main/resources/application.yml
xinhuagu: removed

dev/src/main/resources/application.yml
xinhuagu: I agree. :)  Letting users configure CORS without enabling it by default is safer and more flexible approach. I have removed the application.yml.

dev/src/main/resources/application.yml
vorburger: 👍🏽

dev/src/main/resources/application.yml
vorburger: 👍🏽

PR #307, opened 2025-07-29
Fix missing "model" role in Gemini LLM responses

kaikreuzer:
### Problem

Chat history that was sent on subsequent requests was only storing user messages, missing bot responses due to Gemini class creating Content objects without the required "model" role.

### Solution

Applied fix to all `LlmResponse.create()` calls in Gemini class

### Result

- Bot responses now properly stored in chat history alongside user messages
- Maintains backward compatibility
- All tests pass


Poggecci (approved):
Good catch, thank you!

Issue #302, opened 2025-07-25
[Release Blocker] Dev UI broken on `main` branch (fine on v0.2.0 release)


vorburger:
I've just noticed that the Dev UI is broken on the `main` branch on `cb95b56` (but's it's fine on the last v0.2.0 release, of course). 

Spring Boot cannot find the HTML of the Dev UI anymore, and shows the generic Spring error page; see also #301. (I've not investigated why.)

Noticed this in a project where I'm using the ADK that worked with 0.2.0 and when I switched to using "bleeding edge" `main` by using JitPack (with https://github.com/enola-dev/enola/pull/1565) then it was broken.

This should probably be fixed before the next release. Perhaps a non-regression test for it would be useful as well.

@glaforge @Poggecci @shukladivyansh FYI

vorburger:
> Perhaps a non-regression test for it would be useful as well.

FYI [this](https://github.com/enola-dev/enola/blob/06de584244eec4ca22067abecd692cfd8d621794/java/dev/enola/ai/adk/web/AdkHttpServerTest.java#L27) is what I'm using to detect future regressions in a project using ADK...

... perhaps we should have something like that in ADK itself.

vorburger:
@Poggecci is going to have a closer look into this... Thank You!

vorburger:
@Poggecci FYI https://github.com/enola-dev/enola/pull/1600 is how I can (still, even with today's latest rev) repro this, in my project.

I'll now see if I can also repro it using a simpler basic ADK Java Quickstart like project.

While I'm at it, I'll also look into contributing test coverage for this (if needed).

vorburger:
@Poggecci yeah OK so this is NOT something specific to my Enola.dev project... this is actually very easily reproducible if one tries out (my) https://github.com/enola-dev/LearningADK/tree/main#jitpack at its current `b80aceb` (which [uses](https://github.com/enola-dev/LearningADK/blob/b80aceb1631c2f214905f025aa2c6bd7eb941e3e/jitpack/pom.xml#L33) the current ADK Java [417a8bc](https://github.com/google/adk-java/commit/417a8bcf12bb2c21cf869554335f9649f9ff7a56)).

vorburger:
@Poggecci I believe that I now better understand the root cause of this. I'd like to TDD and get tests in, and then fix it; will contrib.

vorburger:
[...]

PR #342, opened 2025-08-13
test: Introduce AdkWebServerTest

vorburger:
@Poggecci how about this?

This does NOT (yet) cover the #302 problem (if that even is a real problem; TBC) but just tests the ADK Dev Web API - which seems useful in its own right.

I may follow-up on this by adding test coverage for the actual HTML UI, but let's first review this, and I'll do that (or not) in a future separate PR.


Poggecci (approved):
Thank you for adding this! Less likely we'll have silent web server breakages now. Super valuable since we can't automatically test this internally

PR #375, opened 2025-08-19
feat: allow runAsync with null newMessage

wenhaozhao:
feat: allow runAsync with null newMessage


Poggecci:
Thank you for your contribution!

Ooc, what is your usecase for allowing new message to be null?

wenhaozhao:
> Thank you for your contribution!
> 
> Ooc, what is your usecase for allowing new message to be null?

In my project, I need to restore memory content from persistent storage into the session so that the agent can continue execution without the user needing to input a new prompt.

Additionally, this operation was allowed in version 0.2.0, but an incompatibility arose after upgrading to 0.2.1-snapshot.

Poggecci (approved):

Jacksunwei:
@wenhaozhao This is actually not a valid input. The recommended approach is to add a meanful User content to continue the conversation. 

For context, in ADK python, we're introducing "resume" to allow the Runner to resume from a previous incomplete state, but that's also different from taking a null new_message.

wenhaozhao:
> @wenhaozhao This is actually not a valid input. The recommended approach is to add a meanful User content to continue the conversation.
> 
> For context, in ADK python, we're introducing "resume" to allow the Runner to resume from a previous incomplete state, but that's also different from taking a null new_message.

From a logical effect perspective, "resume" is indeed different from a "null-message." However, from a practical effect standpoint, it's similar to: restoring the history to the session, and then by submitting a null-message, causing the runner to resume execution from its previous state.

Currently, adk-java does not have an entry point (or equivalent functionality) similar to "resume." Additionally, "resume" is indeed a better idea.

glaforge:
I'm revisiting some of our open PRs.
Have you found a workaround to this without this PR merged? Or are you handling the situation differently?

wenhaozhao:
Thanks for your reply.

Currently, I'm using a heavily customized ADK implementation branch in my business. However, since the ADK lacks sufficient flexibility for my business needs, we've planned to switch to Spring-ai instead.

Whether this PR's mentioned issue gets merged is no longer relevant now.

ddobrin:
Hi @wenhaozhao 
ADK Java supports an integration with SpringAI at this time. 

Do you wish to discuss in a session which features you'd be looking for and how we could provide assistance?

hemasekhar-p:
Hi @wenhaozhao, could you please provide your thoughts on [this](https://github.com/google/adk-java/pull/375#issuecomment-3986794664)? If you would like to move forward, please share your complete use case in detailed and address the following comments?
2. As per our contribution policy, please squash your changes into a single commit and make sure your PR having only your changes.
3. Could you please address the merge conflicts.

wenhaozhao:
Since ADK doesn't meet my requirements for flexibility, I'm no longer using it for agent development. Currently, I prefer building agent systems in Rust, so this PR can be closed.
Thanks

PR #463, opened 2025-09-28
feat: Add inputAudioTranscription support to Java ADK

jinnigu:
# Summary
Adds `inputAudioTranscription` support to the Java ADK to achieve feature parity with Python.
When enabled, the live connect config requests model-side transcription of input audio into text, allowing real-time conversation processing and downstream analysis.
# Changes
## RunConfig
- Introduce `inputAudioTranscription` field with builder support.
- Preserve existing defaults and semantics.
## Runner
- In live runs with audio modalities, default `inputAudioTranscription` (and `outputAudioTranscription`) when not explicitly set and TEXT is not included, mirroring the Python logic.
## LLM Request Building (Basic)
- Populate `LiveConnectConfig.inputAudioTranscription` from RunConfig.
## Tests
- RunConfigTest: validates setting and default behavior of `inputAudioTranscription`.
- BasicTest: verifies `LiveConnectConfig` is built with `inputAudioTranscription` (and combined with other configs).

### DIFF
--- a/core/src/main/java/com/google/adk/runner/Runner.java
+++ b/core/src/main/java/com/google/adk/runner/Runner.java
@@ -319,9 +319,15 @@ public class Runner {
         if (runConfig.outputAudioTranscription() == null) {
           runConfigBuilder.setOutputAudioTranscription(AudioTranscriptionConfig.builder().build());
         }
+        if (runConfig.inputAudioTranscription() == null) {
+          runConfigBuilder.setInputAudioTranscription(AudioTranscriptionConfig.builder().build());
+        }
       } else if (!runConfig.responseModalities().contains(new Modality(Modality.Known.TEXT))) {
         if (runConfig.outputAudioTranscription() == null) {
           runConfigBuilder.setOutputAudioTranscription(AudioTranscriptionConfig.builder().build());
+        }
+        if (runConfig.inputAudioTranscription() == null) {
+          runConfigBuilder.setInputAudioTranscription(AudioTranscriptionConfig.builder().build());
         }
       }
     }
--- a/core/src/main/java/com/google/adk/agents/RunConfig.java
+++ b/core/src/main/java/com/google/adk/agents/RunConfig.java
@@ -48,7 +48,11 @@ public abstract class RunConfig {
 
   public abstract @Nullable AudioTranscriptionConfig outputAudioTranscription();
 
+  public abstract @Nullable AudioTranscriptionConfig inputAudioTranscription();
+
   public abstract int maxLlmCalls();
+
+  public abstract Builder toBuilder();
 
   public static Builder builder() {
     return new AutoValue_RunConfig.Builder()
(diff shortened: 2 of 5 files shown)


core/src/main/java/com/google/adk/runner/Runner.java
```diff
@@ -366,8 +366,9 @@ private Single<Session> emitStateDeltaEvent(
   private InvocationContext newInvocationContextForLive(
       Session session, Optional<LiveRequestQueue> liveRequestQueue, RunConfig runConfig) {
     RunConfig.Builder runConfigBuilder = RunConfig.builder(runConfig);
-    if (!CollectionUtils.isNullOrEmpty(runConfig.responseModalities())
-        && liveRequestQueue.isPresent()) {
+    if (liveRequestQueue.isPresent() && !this.agent.subAgents().isEmpty()) {
+      // Parity with Python: apply modality defaults and transcription settings
+      // only for multi-agent live scenarios.
```
vorburger, line 402: @jinnigu The inline comment and the code don't seem to align, here? The "text" says "apply modality defaults" but then this removes `!CollectionUtils.isNullOrEmpty(runConfig.responseModalities()`... is that intentional? (It may well be, I'm entirely sure about why this was originally like this; but it seems worth double checking.) Also, why would we limit transcription only for multi-agent live scenarios? I would personally love to use this even for a very simple trivial only-LlmAgent use case... you speak to it, and get a persistent transcript in your session store, that's very cool! I'd love to use this e.g. in my (personal) https://docs.enola.dev project - but don't see why it needs to be limited to work only if `!this.agent.subAgents().isEmpty()`.

vorburger (changes requested):
@jinnigu Thank You for contributing this! Is there any way to illustrate that this actually really 😆  fully works, in this PR? The unit tests are... well, unit tests. Would a full-blown integration test for this be possible? Or, how would you feel about if I invite you to add, as part of this PR, a very (most) simple "MVP" in `tutorials/audio` with just a super simple `LlmAgent` (without even any sub-agents), with just [an `AdkWebServer.start()`](https://github.com/google/adk-java/blob/54bee7ba4de7d59599eaa665659c61e524143efe/tutorials/city-time-weather/src/main/java/com/google/adk/tutorials/CityTimeWeather.java#L99), which allows us to "see this work in action"? That would be awesome!

core/src/main/java/com/google/adk/runner/Runner.java
jinnigu: The reason why I made the change like this is because I want to make adk-java equivalent to adk-python (https://github.com/google/adk-python/blob/main/src/google/adk/runners.py#L939-L971). I also agree that we should not limit transcription to multi-agent live scenarios only. I will raise an issue in adk-python to gather some feedbacks and make PRs to both adk-python and adk-java.

vorburger: @jinnigu awesome! (It's https://github.com/google/adk-python/pull/3259.)

PR #483, opened 2025-10-07
chore: Restore .vscode/settings.json, and document why

vorburger:
This restores #366 which was reverted in #471.

It documents why.


Jacksunwei:
Each developer has their own setup, checking in this would cause headache. It's also not a good practice to checkin settings.json. We can have a recommended settings on the side as newcomers' starting point.

vorburger:
> Each developer has their own setup, checking in this would cause headache. It's also not a good practice to checkin settings.json. We can have a recommended settings on the side as newcomers' starting point.

@Jacksunwei may I respectfully but fundamentally disagree with you about this? Check out the addition to the `CONTRIBUTING.md` in this PR for why... The current `settings.json`, originally proposed in #366 and reverted in #471, are (quite intentionally), very minimal. They are "required", and not just a "recommendation" or "preference". More specifically, they make sure that working with this code in VS Code matches the code format expectations of the Maven build. It is absolutely a good best practice to checkin such basic settings, so that any contributor can be "up and running" immediately with a correctly configured development. Could I cordially invite you to do please elaborate on how they could ever cause any headache to anyone? 

What you mean is that one may also have OTHER (!) VS Code settings which are your "personal preferences" instead of "project requirements". And I totally agree with you about that! But that's something else. Those belong into personal settings in your VS Code User Settings instead. The location of this file varies by platform (e.g., `~/.config/Code/User/settings.json` on Linux). That is precisely why VSC has 2 separate configurations - a workspace (for a project, like this; which we need to have, but keep minimal), and a separate individual personal one.

I hope that this additional context that I've hereby provided helps you to see why this makes sense and is required?

Issue #473, opened 2025-10-05
ADK in general and FunctionTool must allow injecting one's custom ObjectMapper instead of hard-coding it


vorburger:
**Describe the bug**

ADK currently hard-codes `new ObjectMapper().registerModule(new Jdk8Module())` in places like [here](https://github.com/google/adk-java/blob/882d4d9498b82fcbbadb5b234d3f09f0a6001d4f/core/src/main/java/com/google/adk/tools/FunctionTool.java#L46C7-L46C58).

This is well intended ("it works out of the box"), but it's not extensible, and blocks one from using a custom ObjectMapper; e.g. in my project I would like (have) to use [this one instead](https://github.com/enola-dev/enola/blob/21e7d4c650c7f924028cade6a660a7db292e06d5/java/dev/enola/common/io/object/jackson/JacksonObjectReaderWriter.java#L77); notably for:

```java
mapper.registerModule(new JavaTimeModule());
```

But this (the `JavaTimeModule`) is really just one case - the point is that users should be able to inject their own.

The specific issue I've faced is that [my `ToDoTool` here](https://github.com/enola-dev/enola/blob/f95f57696b4c3e06cabc79f83ae6fc0d9491179d/java/dev/enola/todo/ai/tool/adk/ToDoTool.java#L62) used [a domain `class ToDo` with `Instant` fields](https://github.com/enola-dev/enola/blob/f95f57696b4c3e06cabc79f83ae6fc0d9491179d/java/dev/enola/todo/ToDo.java#L39), and that made choke.

I'm temporarily working around it with https://github.com/enola-dev/enola/pull/1855 - but I shouldn't have to.

**To Reproduce**
1. `git clone https://github.com/enola-dev/enola.git`
2. Revert https://github.com/enola-dev/enola/pull/1855
3. Run `b test //java/dev/enola/todo/...`

**Expected behavior**

**Additional context**

@Poggecci @Doris26 @shukladivyansh @Jacksunwei @glaforge agreed that this makes sense?

vorburger:
I've just remember that there is, of course, the [`JsonBaseModel`](https://github.com/google/adk-java/blob/882d4d9498b82fcbbadb5b234d3f09f0a6001d4f/core/src/main/java/com/google/adk/JsonBaseModel.java#L29) already; so I think that this issue is actually 2 somewhat separate things:

[...]

PR #487, opened 2025-10-07
fix: Allow injecting ObjectMapper in FunctionTool, default to ObjectMapper (re. #473)

vorburger:
Related to #473.

Intentionally just for `FunctionTool`, for now; but this should eventually really be done everywhere like this (IMHO).

Alternative (and superior) to the initial #486 - but put on hold, until at least that basic version is merged (thus this is Draft).

### DIFF
--- a/core/src/main/java/com/google/adk/tools/FunctionTool.java
+++ b/core/src/main/java/com/google/adk/tools/FunctionTool.java
@@ -44,12 +44,12 @@ import org.slf4j.LoggerFactory;
 public class FunctionTool extends BaseTool {
 
   private static final Logger logger = LoggerFactory.getLogger(FunctionTool.class);
-  private static final ObjectMapper objectMapper = JsonBaseModel.getMapper();
 
   private final @Nullable Object instance;
   private final Method func;
   private final FunctionDeclaration funcDeclaration;
   private final boolean requireConfirmation;
+  private final ObjectMapper objectMapper;
 
   public static FunctionTool create(Object instance, Method func) {
     return create(instance, func, /* requireConfirmation= */ false);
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/adk/tools/FunctionTool.java
```diff
@@ -305,7 +309,7 @@ public Flowable<Map<String, Object>> callLive(
     }
   }
 
-  private static List<Object> createList(List<Object> values, Class<?> type) {
+  private List<Object> createList(List<Object> values, Class<?> type) {
```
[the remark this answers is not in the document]
vorburger: ==> #489

core/src/main/java/com/google/adk/tools/FunctionTool.java
```diff
@@ -144,6 +149,7 @@ protected FunctionTool(@Nullable Object instance, Method func, boolean isLongRun
     this.funcDeclaration =
         FunctionCallingUtils.buildFunctionDeclaration(
             this.func, ImmutableList.of("toolContext", "inputStream"));
+    this.objectMapper = objectMapper;
```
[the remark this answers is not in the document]
vorburger: TBD...

glaforge:
Looks like the code is conflicting with the current version of the codebase.
Would need to be updated and the merge conflicts be resolved.

vorburger:
> Looks like the code is conflicting with the current version of the codebase. Would need to be updated and the merge conflicts be resolved.

@glaforge I'll take care of this...

core/src/main/java/com/google/adk/tools/FunctionTool.java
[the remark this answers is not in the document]
vorburger: done

core/src/main/java/com/google/adk/tools/FunctionTool.java
```diff
@@ -44,12 +44,13 @@
 public class FunctionTool extends BaseTool {
 
   private static final Logger logger = LoggerFactory.getLogger(FunctionTool.class);
-  private static final ObjectMapper objectMapper = JsonBaseModel.getMapper();
+  private static final ObjectMapper defaultObjectMapper = JsonBaseModel.getMapper();
```
krwc: Can we remove this completely? It seems objectMapper provided in the constructor is always in use in this class?

vorburger: > It seems objectMapper provided in the constructor is always in use in this class?

@krwc that not entirely accurate, with the PR (as-is), the existing constructors are kept (intentionally), and it "falls back" to that `defaultObjectMapper = JsonBaseModel.getMapper();`.

> Can we remove this completely? 

Only if we agree to break the public API surface of the `FunctionTool`, and change its (`protected`, but still) constructor signatures. Personally I would not do that, as there really is no need - we can easily keep this API backwards compatible (with the PR as-is) and default to that current original mapper - yet allow injecting a custom one.

But one thing that I probably could do, without changing above, just as a "clean up", which might (?) make this more obvious to future readers of this code, is to "inline" the `JsonBaseModel.getMapper())` to the existing constructors, instead of keeping it in a `private static final`... would you like me to do that?

krwc: Ack - I missed that. I'd consider inlining `JsonBaseModel.getMapper()` where appropriate, other than that LGTM.

vorburger: > consider inlining `JsonBaseModel.getMapper()` where appropriate, other than that LGTM.

@krwc done! PTAL?

Issue #504, opened 2025-10-15
MCP servers execution result doesn't sent back to model


nikoncode:
**Describe the bug**
MCP servers execution result doesn't sent back to model (with Claude).

**To Reproduce**
1. Open McpFilesystemAgent example from repository
2. Configure it to use Claude model
3. Execute

**Expected behavior**
List of files returned to user

**Actual behavior**
MCP calls executed successfully, but the results weren’t sent to the model.
```
You> List the files available in /Users/xxx/test
Agent> I'll help you list the files in the specified directory. Let me check what's available in that location.Function Call: FunctionCall{id=Optional[toolu_01FMMHBMJ1vWMVHyaFq3tWXt], args=Optional[{path=/Users/xxx/test}], name=Optional[list_directory]}Function Response: FunctionResponse{willContinue=Optional.empty, scheduling=Optional.empty, parts=Optional.empty, id=Optional[toolu_01FMMHBMJ1vWMVHyaFq3tWXt], name=Optional[list_directory], response=Optional[{text_output=[{text=[FILE] 1.txt
        [FILE] 2.txt}]}]}I don't see any output from the listing command. Let me first check which directories are actually allowed for access in the system:Function Call: FunctionCall{id=Optional[toolu_016E5pBJhUFTXtEzp9wWVHoQ], args=Optional[{}], name=Optional[list_allowed_directories]}Function Response: FunctionResponse{willContinue=Optional.empty, scheduling=Optional.empty, parts=Optional.empty, id=Optional[toolu_016E5pBJhUFTXtEzp9wWVHoQ], name=Optional[list_allowed_directories], response=Optional[{text_output=[{text=Allowed directories:
        /Users/xxx/test}]}]}It seems I'm unable to access the list of allowed directories or the specific path you requested. This could be due to one of the following reasons:

        1. The directory "/Users/xxx/test" doesn't exist
        2. The path is outside the allowed directories that this assistant can access
3. There might be permission issues with the directory

Could you please:
        1. Confirm if the path is correct?
        2. Try providing a different directory path that you think should be accessible?
        3. If you're unsure which directories are available, you might need to check with the system administrator about which directories are allowed for this assistant to access.
```

**Desktop (please complete the following information):**
 - OS: MacOS
 - Java version: 17
 - ADK version(see maven dependency): 0.3.0

**Possible reason:**
spent some time troubleshooting this issue. It occurs because the function call response contains data under the `text_output` key, while Claude expects it under the `result` key.
Expected:
[...]

PR #505, opened 2025-10-15
Let the MCP tool execution result be sent back to the model

nikoncode:
Very naive way to handle #504

### DIFF
--- a/core/src/main/java/com/google/adk/tools/mcp/AbstractMcpTool.java
+++ b/core/src/main/java/com/google/adk/tools/mcp/AbstractMcpTool.java
@@ -161,6 +161,6 @@ public abstract class AbstractMcpTool<T> extends BaseTool {
         resultMaps.add(ImmutableMap.of("text", textOutput));
       }
     }
-    return ImmutableMap.of("text_output", resultMaps);
+    return ImmutableMap.of("result", resultMaps);
   }
 }


shukladivyansh:
Hi Mikita!

Thanks for looking into this. Can you do the fix in Claude.java instead ([adk-java/core/src/main/java/com/google/adk/models/Claude.java](https://github.com/google/adk-java/blob/939de25223b854293c001e81cd4cb02c0070b091/core/src/main/java/com/google/adk/models/Claude.java#L174)) since others models may be depending on the current implementation of [core/src/main/java/com/google/adk/tools/mcp/AbstractMcpTool.java](https://github.com/google/adk-java/pull/505/files#diff-c777e47d8db7e7d20ff582eaad697a9f888e89a1cfa74ef67e98ab4cb96a67f5)

nikoncode:
@shukladivyansh for sure, but:

* I was not able to find any usage
* I checked python adk and they have same logic relying on `result` field in Claude implementation

That's why I decided to proceed with fix in McpTool. Also I don't have keys to other models to verify that they work fine.

shukladivyansh:
>I was not able to find any usage

Yes, in this codebase that is. Users downstream might be reusing our tools/models and it would break them. 

> I checked python adk and they have same logic relying on result field in Claude implementation

I agree that we plan to follow adk python as closely as possible. But breaking users doesn't feel worth it in this case. Feel free to add a comment there that this logic deviates from adk python.

> That's why I decided to proceed with fix in McpTool. Also I don't have keys to other models to verify that they work fine.

Totally understandable. 

I would insist in this case to proceed with a minimal fix in Claude llm implementation. This wouldn't break users and also allow you to verify things on your end.

core/src/main/java/com/google/adk/tools/mcp/AbstractMcpTool.java
```diff
@@ -152,6 +152,6 @@ protected static Map<String, Object> wrapCallResult(
         resultMaps.add(ImmutableMap.of("text", textOutput));
       }
     }
-    return ImmutableMap.of("text_output", resultMaps);
```
Poggecci, line 164: Would you mind verifying that this change doesn't break the functionality for non-Claude models?

krwc:
Extensive fix in https://github.com/google/adk-java/pull/1078.

Issue #510, opened 2025-10-16
Gemini thoughts not correctly accumulated when streaming enabled


OwenDavisBC:
**Description**
When using gemini models with thinking enabled and thoughts turned on, the thoughts are being incorrectly appended to non-thought text responses from gemini when streaming is enabled. The python-adk does not exhibit the same behavior of combing thoughts and non-thought text responses.

The python-adk explicitly builds thought text responses separately from non-thought response to avoid this issue. See - https://github.com/google/adk-python/blob/main/src/google/adk/utils/streaming_utils.py#L51-L82

It can also be seen in the python-adk, when building llm requests, that thoughts are not removed. See the python-adk llm request pre-processing - https://github.com/google/adk-python/blob/main/src/google/adk/models/google_llm.py#L291

**To Reproduce**
1. Setup an agent with thinking enabled, for example the hello world agent:
```java
LlmAgent.builder()
  .name("hello_world_agent")
  .description("hello world agent that can roll a dice and check prime numbers.")
  .model("gemini-2.5-flash")
  .instruction(
      """
      You roll dice and answer questions about the outcome of the dice rolls.
      You can roll dice of different sizes.
      You can use multiple tools in parallel by calling functions in parallel(in one request and in one round).
      It is ok to discuss previous dice roles, and comment on the dice rolls.
      When you are asked to roll a die, you must call the roll_die tool with the number of sides. Be sure to pass in an integer. Do not pass in a string.
      You should never roll a die on your own.
      When checking prime numbers, call the check_prime tool with a list of integers. Be sure to pass in a list of integers. You should never pass in a string.
      You should not check prime numbers before calling the tool.
      When you are asked to roll a die and check prime numbers, you should always make the following two function calls:
      1. You should first call the roll_die tool to get a roll. Wait for the function response before calling the check_prime tool.
      2. After you get the function response from roll_die tool, you should call the check_prime tool with the roll_die result.
        2.1 If user asks you to check primes based on previous rolls, make sure you include the previous rolls in the list.
      3. When you respond, you must include the roll_die result from step 1.
      You should always perform the previous 3 steps when asking for a roll and checking prime numbers.
      You should not rely on the previous history on prime results.
      """)
  .tools(
      ImmutableList.of(
          FunctionTool.create(HelloWorldAgent.class, "rollDie"),
          FunctionTool.create(HelloWorldAgent.class, "checkPrime")))
  .generateContentConfig(
      GenerateContentConfig.builder()
          .safetySettings(
[...]

PR #514, opened 2025-10-17
fix: Gemini thoughts not correctly accumulated when streaming enabled

OwenDavisBC:
Fixes https://github.com/google/adk-java/issues/510

### DIFF
--- a/core/src/main/java/com/google/adk/models/Gemini.java
+++ b/core/src/main/java/com/google/adk/models/Gemini.java
@@ -35,6 +35,7 @@ import io.reactivex.rxjava3.core.Flowable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -236,6 +237,7 @@ public class Gemini extends BaseLlm {
 
   static Flowable<LlmResponse> processRawResponses(Flowable<GenerateContentResponse> rawResponses) {
     final StringBuilder accumulatedText = new StringBuilder();
+    final StringBuilder accumulatedThoughtText = new StringBuilder();
     // Array to bypass final local variable reassignment in lambda.
     final GenerateContentResponse[] lastRawResponseHolder = {null};
     return rawResponses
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/google/adk/models/Gemini.java
```diff
@@ -246,15 +248,26 @@ static Flowable<LlmResponse> processRawResponses(Flowable<GenerateContentRespons
...
+                } else {
+                  accumulatedText.append(currentTextChunk);
+                }
                 LlmResponse partialResponse =
                     currentProcessedLlmResponse.toBuilder().partial(true).build();
                 responsesToEmit.add(partialResponse);
               } else {
+                if (accumulatedThoughtText.length() > 0
+                    && GeminiUtil.shouldEmitAccumulatedText(currentProcessedLlmResponse)) {
+                  LlmResponse aggregatedTextResponse =
```
Poggecci: aggregatedThoughtResponse here?

Poggecci (changes requested):
Hey Owen. Thank you for the contribution!

This PR seems to both resolve the thoughts accumulation issue, but also changes what content we send in our requests to the model. Is there a reason you've coupled these two changes? The original rationale behind stripping thoughts was a naive form of context management (although the relevance of this is fully dependent on the GenAI SDK's handling of the thoughts we provide), but since we've had this behavior since release, a change to it merits its own PR. 

Happy to approve if just the accumulation is kept in this PR or if we have some further discussion on the matter.

OwenDavisBC:
@Poggecci thank you for taking a look. The issue with stripping thoughts is that will ultimately lead to empty parts being sent to the gemini API now that we are accumulating thought-only parts. I can add some more processing to remove those from the llm request here https://github.com/google/adk-java/blob/7f1206402fe97081ccc9aaf661f8a513c3b25ce4/core/src/main/java/com/google/adk/models/Gemini.java#L208-L235, but that would not match what I see in the python-adk. 

In the python-adk if I look at pre-processing before https://github.com/google/adk-python/blob/4a842c5a1334c3ee01406f796651299589fe12ab/src/google/adk/models/google_llm.py#L149-L154
```python
    if stream:
      responses = await self.api_client.aio.models.generate_content_stream(
          model=llm_request.model,
          contents=llm_request.contents,
          config=llm_request.config,
      )
```
I see nothing removing thoughts - https://github.com/google/adk-python/blob/4a842c5a1334c3ee01406f796651299589fe12ab/src/google/adk/models/google_llm.py#L300-L326.
```python
  async def _preprocess_request(self, llm_request: LlmRequest) -> None:

    if self._api_backend == GoogleLLMVariant.GEMINI_API:
      # Using API key from Google AI Studio to call model doesn't support labels.
      if llm_request.config:
        llm_request.config.labels = None

      if llm_request.contents:
        for content in llm_request.contents:
          if not content.parts:
            continue
          for part in content.parts:
            # Create copies to avoid mutating the original objects
            if part.inline_data:
              part.inline_data = copy.copy(part.inline_data)
              _remove_display_name_if_present(part.inline_data)
            if part.file_data:
              part.file_data = copy.copy(part.file_data)
              _remove_display_name_if_present(part.file_data)

    # Initialize config if needed
    if llm_request.config and llm_request.config.tools:
      # Check if computer use is configured
      for tool in llm_request.config.tools:
        if isinstance(tool, types.Tool) and tool.computer_use:
          llm_request.config.system_instruction = None
          await self._adapt_computer_use_tool(llm_request)
```

OwenDavisBC:
Hi @Poggecci, let me know how you and folks are feeling about the change as is. I am open to omitting the thoughts to save tokens, I will just need the extra filter to remove empty parts from llm requests.

Poggecci:
Thanks  for following up Owen! I did some digging on GenAI SDK and persisting thoughts is configurable through separate params soI am good with not stripping them anymore.

PR #537, opened 2025-10-21
fix: listSessions returns sessions with empty state

tomekpanek:
May look like optimization but [python version ](https://github.com/google/adk-python/blob/fe1fc75c15a7983829bbe0b023f4b612b1e5c018/src/google/adk/sessions/vertex_ai_session_service.py#L240)returns non-empty session states.


Poggecci:
Please  update your base branch and we can get this change checked in. Thank you!

tomekpanek:
> Please update your base branch and we can get this change checked in. Thank you!

With some problems but updated now.

PR #577, opened 2025-11-10
Make AgentStaticLoader public

kawolum:
This is needed to start an application with custom Spring configuration. AgentStaticLoader will be instantiated as a bean.

### DIFF
--- a/dev/src/main/java/com/google/adk/web/AgentStaticLoader.java
+++ b/dev/src/main/java/com/google/adk/web/AgentStaticLoader.java
@@ -37,7 +37,7 @@ import javax.annotation.Nonnull;
  * <p>This class is not a Spring component by itself - instances are created programmatically and
  * then registered as beans via factory methods.
  */
-class AgentStaticLoader implements AgentLoader {
+public class AgentStaticLoader implements AgentLoader {
 
   private final ImmutableMap<String, BaseAgent> agents;
 


vorburger:
@kawolum seeing this PR made me laugh 🤣 and chuckle, and made my day, because we've gone back and forth:

* In https://github.com/kawolum/adk-java/commit/5fcd4136aadb6e3592c643d766e3ea9d8917cb41#diff-4a4d4aff2de61b4c7f45bed057a53f9796966da6e0b253a395cfa69e46abbe00, I introduced it as package-private

* In https://github.com/kawolum/adk-java/commit/0f7904b903095cfe38822b70d608b4c6899b667a#diff-4a4d4aff2de61b4c7f45bed057a53f9796966da6e0b253a395cfa69e46abbe00, @Doris26 made it `public` as part of a great refactoring that lead to a much better design than my original stab.

* In https://github.com/kawolum/adk-java/commit/87acdf8083e7c4dc92273dcde0048ee9be2882fd#diff-4a4d4aff2de61b4c7f45bed057a53f9796966da6e0b253a395cfa69e46abbe00, @Doris26 made the `public` again `package-private`, presumably because _"This class is not a Spring component by itself - instances are created programmatically and then registered as beans via factory methods."_

* Here it's being proposed to made `public` again! 🤣 

I'm OK with making it `public` after all. @Doris26 do you have any pro for or con against this?

vorburger:
> I'm OK with making it `public` after all. @Doris26 do you have any pro for or con against this?

Having said that... @kawolum can you elaborate a little further on why you need it?

> This is needed to start an application with custom Spring configuration. AgentStaticLoader will be instantiated as a bean.

Do you just want to use `AgentStaticLoader` as is, in a Spring configuration class? Would making it `final` be OK for you?

Or do you want to sub-class and change it? Then wouldn't it be simpler to just implement `AgentLoader` yourself?

kawolum:
I am deploying ADK web as a service and I need to implement a `/healthcheck` path.
I have my own Spring Configuration that adds the health check controller bean and `AgentLoader` bean. (I also need to add some other beans.)
I am doing something like this to start the server.
```
SpringApplication app = new SpringApplication(AdkWebServer.class, MyConfiguration.class);
app.run(args);
```

Okay with making `AgentStaticLoader` final, but I didn't initially because `CompiledAgentLoader` is not. I am okay either way.
I currently have a copy of `AgentStaticLoader` since the current implementation is perfect for my use case.

kawolum:
@vorburger Let me know what you think. I can clean up the commit in the PR afterwards.

mazas-google (approved):

mazas-google:
Looks good for merging. Could you squash and rebase?

kawolum:
Yup rebased! @mazas-google

mazas-google:
Needs another rebase, branch is out-of-date :(

mazas-google:
To speed things up I submitted this change in https://github.com/google/adk-java/commit/cf5be2537431179465348091248c457102cb8e23, because I need to cut 0.4.0 release.

PR #609, opened 2025-11-22
Fix tool result processing in Claude model

### LINKED ISSUE #566: Result of tool call ignored if the result is not put under the "result" key
Image

**Describe the bug**
I am invoking a tool and the result is ignored except if I wrap it in the "result" key of the returned map.

Using this:

```groovy
dependencies {
    implementation(group: 'com.google.adk', name: 'google-adk-dev', version: '0.3.0')
    implementation(group: 'com.google.adk', name: 'google-adk', version: '0.3.0')
}
```

**To Reproduce**

With the `return Map.of("count", 123L)`, the agent does not pick up the result.
If i replace it with `Map.of("result", 123L)`, all works well.

```java
public class AdkPOC {

    private static final String CLAUDE_MODEL_ID = "claude-haiku-4-5";

    public static void main(String[] args) {

        var anthropicClient = AnthropicOkHttpClient.builder()
                .apiKey(System.getenv("ANTHROPIC_API_KEY"))
                .build();

        var claudeModel = new Claude(
                CLAUDE_MODEL_ID,
                anthropicClient
        );

        var dataAnalyst = LlmAgent.builder()
                .name("Data Analyst")
                .description("An agent to perform data analytics with multiple tools")
                .model(claudeModel)
                .tools(FunctionTool.create(AdkPOC.class, "getRowCount"))
                .build();

        AdkWebServer.start(dataAnalyst);
    }

    public static Map<String, Object> getRowCount(
            @Schema(name = "dataSetName", description = "The name of the dataset to query")
            String sheetName
    ) {
        return Map.of("count", 123L);
    }
}
```

**Expected behavior**

I would expect that the structure of the resulting map is not imposed

stevearmstrong-dev:
Fixes #566

## Problem
Tool results were only processed if they contained a 'result' key, causing other valid Map structures to be ignored.

## Solution  
Changed the Claude model to serialize the entire response Map to JSON instead of extracting only the 'result' key.

## Changes
- Modified `Claude.java:171-182` to use `ObjectMapper.writeValueAsString()` 
- Added proper error handling with try-catch
- Falls back to `.toString()` if JSON serialization fails

## Testing
- ✅ All FunctionTool tests pass (36/36)
- ✅ Code compiles successfully
- ✅ Verified with the reproduction case from the issue

### DIFF
--- a/a2a/README.md
+++ b/a2a/README.md
@@ -177,7 +177,7 @@ curl -X POST http://localhost:8081/a2a/remote/v1/message:send \
             "kind": "message",
             "contextId": "cli-demo-context",
             "messageId": "cli-check-2",
-            "role": "USER",
+            "role": "user",
             "parts": [
               { "kind": "text", "text": "Is 2 prime?" }
             ]
--- a/a2a/pom.xml
+++ b/a2a/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>com.google.adk</groupId>
     <artifactId>google-adk-parent</artifactId>
-    <version>0.3.1-SNAPSHOT</version><!-- {x-version-update:google-adk:current} -->
+    <version>0.4.1-SNAPSHOT</version><!-- {x-version-update:google-adk:current} -->
   </parent>
 
   <artifactId>google-adk-a2a</artifactId>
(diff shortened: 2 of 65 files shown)


stevearmstrong-dev:
@mazas-google @Poggecci Could you please review this PR?

krwc:
Addressed to align with Python ADK logic in https://github.com/google/adk-java/pull/1078.

Issue #559, opened 2025-10-29
Parallel Agent not executing sub agents concurrently by default.


ankity:
** Please make sure you read the contribution guide and file the issues in the rigth place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**
Google ADK's `ParallelAgent` (v0.3.0) does **NOT** execute sub-agents in parallel. All agents run sequentially on the same thread.

**To Reproduce**
**Simple test agent with delay:**
```java
public class DelayedTestAgent extends BaseAgent {
  private final long delayMillis;

  public DelayedTestAgent(String name, long delayMillis) {
    super(name, "Test agent with " + delayMillis + "ms delay", null, null, null);
    this.delayMillis = delayMillis;
  }

  @Override
  public Flowable<Event> runAsyncImpl(InvocationContext context) {
    long startTime = System.currentTimeMillis();
    String threadName = Thread.currentThread().getName();
    
    System.out.println(String.format("[%s] Starting on thread: %s", name(), threadName));
    
    try {
      Thread.sleep(delayMillis);
    } catch (InterruptedException e) {
      Thread.currentThread().interrupt();
    }
    
    long duration = System.currentTimeMillis() - startTime;
    System.out.println(String.format("[%s] Completed in %dms on thread: %s", 
        name(), duration, threadName));
    
    Content content = Content.fromParts(Part.fromText("Done"));
    return Flowable.just(Event.builder().author(name()).content(content).build());
  }

  @Override
  protected Flowable<Event> runLiveImpl(InvocationContext context) {
    throw new UnsupportedOperationException("Not supported");
  }
}
```

**Test:**
```java
@Test
void testParallelAgentExecutesAgentsConcurrently() {
    // Create two agents with 2-second delays each
    DelayedTestAgent agent1 = new DelayedTestAgent("agent-1", 2000);
    DelayedTestAgent agent2 = new DelayedTestAgent("agent-2", 2000);

    // Create ParallelAgent
    ParallelAgent parallelAgent =
        ParallelAgent.builder()
            .name("parallel-test")
            .subAgents(agent1, agent2)
            .build();

    // Execute
    long startTime = System.currentTimeMillis();
    
    InMemoryRunner runner = new InMemoryRunner(parallelAgent);
    Session session = runner.sessionService()
        .createSession("parallel-test", "test-user").blockingGet();
    
    Content userMsg = Content.fromParts(Part.fromText("test"));
[...]

PR #610, opened 2025-11-22
Fix: Parallel Agent not executing sub agents concurrently by default

jinnigu:
Fixes https://github.com/google/adk-java/issues/559

## Problem
Google ADK's ParallelAgent (v0.3.0) was executing sub-agents sequentially on the same thread instead of concurrently. This meant that if one agent blocked or took time, the others need to wait, which negates parallelism performance.

## Changes

- Modified ParallelAgent.java to wrap sub-agent execution with .subscribeOn().
- Added BlockingAgent helper class to ParallelAgentTest.java to simulate long-running tasks.
- Added regression test runAsync_blockingSubAgents_shouldExecuteInParallel to verify parallel execution timing.

## Testing
- ParallelAgentTest passes (including new regression test)
- LlmAgentTest passes (no regressions)
- Verified parallel execution: Test with two 1s blocking agents completes in ~1s instead of 2s.

### DIFF
--- a/core/src/main/java/com/google/adk/agents/BaseAgent.java
+++ b/core/src/main/java/com/google/adk/agents/BaseAgent.java
@@ -285,12 +285,10 @@ public abstract class BaseAgent {
     InvocationContext.Builder builder = parentContext.toBuilder();
     builder.agent(this);
     // Check for branch to be truthy (not None, not empty string),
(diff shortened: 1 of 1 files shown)


laxman-munigala:
Do we know when is this going to be included in the release. This seems one of the basic capability , have to spend lot of time to identify this is a bug/issue with the framework.

krwc:
We've addressed the issue in https://github.com/google/adk-java/pull/1077.

Issue #561, opened 2025-10-30
Support Sub-agent Escalation event in Parallel Agent


harveyatwal-gif:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
If a sub-agent sends a termination signal via an escalation event then short-circuit the concurrent requests.

**Describe alternatives you've considered**
N/A

**Additional context**
The idea is similar to how sub-agents can signal to LoopAgent to stop the loop.

PR #612, opened 2025-11-24
feat: Support Sub-agent Escalation event in Parallel Agent

jinnigu:
Fixes https://github.com/google/adk-java/issues/561

## Problem
ADK ParallelAgent did not support short-circuiting concurrent requests when a sub-agent sent a termination signal via an escalation event.

## Changes
Modified ParallelAgent.java to use .takeUntil() on the merged event stream, listening for events with the escalate flag.
Added ParallelAgentEscalationTest.java to verify the escalation event triggers termination of other sub-agents.

## Testing
- ParallelAgentEscalationTest passes (verifies short-circuit behavior).
- ParallelAgentTest passes (no regressions for normal parallel execution).
- Verified that when one sub-agent escalates, other long-running sub-agents are cancelled immediately.

### DIFF
--- a/core/src/test/java/com/google/adk/agents/ParallelAgentEscalationTest.java
+++ b/core/src/test/java/com/google/adk/agents/ParallelAgentEscalationTest.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.adk.agents;
+
+import static com.google.adk.testing.TestUtils.createInvocationContext;
+import static com.google.common.truth.Truth.assertThat;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+
+import com.google.adk.events.Event;
+import com.google.adk.events.EventActions;
+import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/google/adk/agents/ParallelAgentEscalationTest.java
```diff
@@ -0,0 +1,139 @@
...
+      return event;
+    }
+
+    @Override
+    protected Flowable<Event> runLiveImpl(InvocationContext invocationContext) {
+      throw new UnsupportedOperationException("Not implemented");
+    }
+  }
+
+  static class SlowAgent extends BaseAgent {
```
tilgalas: maybe we could merge the Escalating and Slow agents into a single TestAgent class, I feel like the code only differs in how the Event is constructed in the runAsyncImpl  - this could be specified via constructor params for example, wdyt?

core/src/test/java/com/google/adk/agents/ParallelAgentEscalationTest.java
```diff
@@ -0,0 +1,139 @@
...
+    protected Flowable<Event> runLiveImpl(InvocationContext invocationContext) {
+      throw new UnsupportedOperationException("Not implemented");
+    }
+  }
+
+  @Test
+  public void runAsync_escalationEvent_shortCircuitsOtherAgents() {
+    TestScheduler testScheduler = new TestScheduler();
+
+    EscalatingAgent agent1 = new EscalatingAgent("agent1", 100, testScheduler);
```
tilgalas: can we add another SlowAgent here (which would actually be a fast one) with a delay of say 50 (so that it's completed before the escalation happens) - I wanted to make sure that the EscalatingAgent escalation still preserves the events that happend prior to it completing

core/src/test/java/com/google/adk/agents/ParallelAgentEscalationTest.java
jinnigu: Yes, merged into a single TestAgent class

core/src/test/java/com/google/adk/agents/ParallelAgentEscalationTest.java
jinnigu: Added a third agent to complete before escalation.

PR #626, opened 2025-11-27
feat: add missing tool resolution strategy

tkroman:
hi, I'm not sure if I had to create an issue first so please let me know if this is ok.

the use-case is recovering from exceptions thrown as a result of hallucinated tool calls. i tried to preserve behavior by using the current throwing behaviour wherever this new argument is passed.

also, some tests fail locally for me, i'll try to figure it out but hopefully ci passes

### DIFF
--- a/core/src/main/java/com/google/adk/flows/llmflows/Functions.java
+++ b/core/src/main/java/com/google/adk/flows/llmflows/Functions.java
@@ -30,9 +30,9 @@ import com.google.adk.events.Event;
 import com.google.adk.events.EventActions;
 import com.google.adk.tools.BaseTool;
 import com.google.adk.tools.FunctionTool;
+import com.google.adk.tools.MissingToolResolutionStrategy;
 import com.google.adk.tools.ToolConfirmation;
 import com.google.adk.tools.ToolContext;
-import com.google.common.base.VerifyException;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.genai.types.Content;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/google/adk/flows/llmflows/Functions.java
```diff
@@ -72,6 +72,84 @@ public static String generateClientFunctionCallId() {
...
+   * @param invocationContext The invocation context.
+   * @param validCalls The list of valid function calls.
+   * @param missingToolsFlowable The flowable for missing tool events.
+   * @param functionCallMapper The mapper to convert function calls to events.
+   * @return A combined flowable of all events.
+   */
+  private static Flowable<Event> createCombinedFlowable(
+      InvocationContext invocationContext,
+      ImmutableList<FunctionCall> validCalls,
+      Flowable<Event> missingToolsFlowable,
```
tilgalas, line 139: I feel like `validCalls` and `missingToolsFlowable` can be replaced with a single ToolCallSeparation parameter

core/src/main/java/com/google/adk/tools/MissingToolResolutionStrategy.java
```diff
@@ -0,0 +1,39 @@
...
+  public static final MissingToolResolutionStrategy RETURN_ERROR =
+      (invocationContext, functionCall) ->
+          Maybe.error(
+              new VerifyException(
+                  "Tool not found: " + functionCall.name().orElse(functionCall.toJson())));
+
+  public static final MissingToolResolutionStrategy IGNORE =
+      (invocationContext, functionCall) -> Maybe.empty();
+
+  public static MissingToolResolutionStrategy respondWithEvent(
```
tilgalas, line 26: I think we can get rid of this helper, it's practically a noop

tilgalas (commented):
Hi @tkroman, thanks for your PR - it looks good to me with minor suggestions - you'll probably need to resolve merge conflicts before we can merge it, though

PR #682, opened 2025-12-18
fix: Merging of events in rearrangeEventsForAsyncFunctionResponsesInHistory

r4inee:
**Problem** 

When using long running tools https://google.github.io/adk-docs/tools-custom/function-tools/#long-run-tool (or boolean confirmation), if a follow up prompt is provided to the agent after a "confirmed" tool is executed. The `rearrangeEventsForAsyncFunctionResponsesInHistory` function will incorrectly merge function calls and responses in the event history.

This results in events being not in order, and multiple function response for the same function call, and results in failures in subsequent invocation of the agent when using non-Gemini models.

```
Role 'tool' must be a response to a preceding message with 'tool_calls'
```

**Example**
Input:
[user prompt 1, FC1, user prompt 2, FR1A, user prompt 3, model response 1, FR1B, user prompt 4]

Actual Output:
[user prompt 1, FC1, FR1B, user prompt 2, FR1A, user prompt 3, model response 1, user prompt 4]

My Expected Output:
[user prompt 1, FC1, FR1B, user prompt 4] (dropping all events between the FC and corresponding FR)

Looking at how we handle the merging of events when the function response is the latest event `rearrangeEventsForLatestFunctionResponse`, we seem to drop events between the FC and corresponding FR, perhaps the logic should be smarter to conditionally keep some the intermediate events.

adk-python Output
[user prompt 1, FC1, FR1B, model response 1, user prompt 4]

`adk-python` seems to keep some of the intermediate model responses where the model asks user for confirmation, but providing this to subsequent agent calls doesn't seem very useful if function has already been executed and we have the result.
Screenshot 2026-01-19 at 12 35 26 pm

**Change**
When processing each event, it will skip to the corresponding function response and thus ignoring events in between. Added some tests to handle when there are multiple function calls.

### DIFF
--- a/core/src/main/java/com/google/adk/flows/llmflows/Contents.java
+++ b/core/src/main/java/com/google/adk/flows/llmflows/Contents.java
@@ -564,8 +564,7 @@ public final class Contents implements RequestProcessor {
     for (int i = 0; i < events.size(); i++) {
       Event event = events.get(i);
 
-      // Skip response events that will be processed via responseEventsBuffer
-      if (processedResponseIndices.contains(i)) {
+      if (!event.functionResponses().isEmpty()) {
         continue;
       }
 
--- a/core/src/test/java/com/google/adk/flows/llmflows/ContentsTest.java
+++ b/core/src/test/java/com/google/adk/flows/llmflows/ContentsTest.java
@@ -203,9 +203,11 @@ public final class ContentsTest {
   public void rearrangeHistory_multipleFRsForSameFC_returnsMergedFR() {
     Event fcEvent = createFunctionCallEvent("fc1", "tool1", "call1");
     Event frEvent1 =
-        createFunctionResponseEvent("fr1", "tool1", "call1", ImmutableMap.of("status", "running"));
+        createFunctionResponseEvent("fr1", "tool1", "call1", ImmutableMap.of("status", "pending"));
     Event frEvent2 =
-        createFunctionResponseEvent("fr2", "tool1", "call1", ImmutableMap.of("status", "done"));
+        createFunctionResponseEvent("fr2", "tool1", "call1", ImmutableMap.of("status", "running"));
+    Event frEvent3 =
+        createFunctionResponseEvent("fr3", "tool1", "call1", ImmutableMap.of("status", "done"));
     ImmutableList<Event> inputEvents =
         ImmutableList.of(
             createUserEvent("u1", "Query"),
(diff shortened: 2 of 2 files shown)


r4inee:
👋  @krwc @mazas-google Would you be able to give this a review too when you have a chance! Happy to make any changes  here if neccesary.

r4inee:
Hey @krwc @mazas-google - I have done a bit of testing and a fix here is still required to address the confirmation problem for other LLM models.

Let me know if there is a more appropriate place to address this.

tilgalas:
Hi @r4inee, thanks for your contribution, it looks good to me, but I'm somewhat concerned about the mismatch between the python and java implementations which we'll still have even with your change in place. Let me consult with python maintainers about your proposition and the criteria for keeping or filtering out the intermediate events, and I'll get back to you once I know their recommendation. In the meantime, if you'd like to try to change your logic so that the python and java adks behave the same way, that would help remove any concerns that we have against merging the PR, and we can still later revisit the question of which events should we actually remove or keep.

r4inee:
Hi @tilgalas, thanks for the review - when I get a chance I can try to replicate the adk-python logic in a seperate branch to see what that would look like. But am keen to hear from the python folks and do the right thing here rather than just merging a quick fix.

For reference [this is corresponding logic I have been looking at in adk-python](https://github.com/google/adk-python/blob/1063fa532cad59d8e9f7421ce2f523724d49d541/src/google/adk/flows/llm_flows/contents.py#L122), which adds the non-function-response events after merging/moving FR events to be directly after FC. My main concern with following that implementation is that the merging logic when the confirmation event is the latest event ([_rearrange_events_for_latest_function_response](https://github.com/google/adk-python/blob/1063fa532cad59d8e9f7421ce2f523724d49d541/src/google/adk/flows/llm_flows/contents.py#L127)) is now different to the merging logic when confirmation event is in the previous turns, and the existence of those intermediate events may lead to inconsistent models behaviour, perhaps some models would be smart enough to ignore them.

tilgalas:
Hi @r4inee, I investigated the adk python behavior more closely and I have a couple of thoughts:
1. The adk_python output doesn't actually remove any events, just pushes them outside the (function call, function response) pair - so I believe, with your input you should be observing output like this:
    ```
    [user prompt 1, FC1, FR1B, user prompt 2, user prompt 3, model response 1, user prompt 4]
    ```
    in other words - only the FR1A event has been filtered out, since it's been "superseeded" by FR1B, everything else stays as it was, modulo being shifted outside the function call-response bracket
2. It seems like this is indeed a bug in java's implementation, in particular the code [here](https://github.com/google/adk-python/blob/1063fa532cad59d8e9f7421ce2f523724d49d541/src/google/adk/flows/llm_flows/contents.py#L95-L97) explictly filters out any response seen, and as the comment explains only "pastes" them back in the function call processing in the subsequent branch
3. As mentioned above, Java code seems to be lacking this filtering out of response events, would you like to try to update your PR to try to mimic that behavior? 
4. It looks like the removal of all events between the call and response in [_rearrange_events_for_latest_function_response](https://github.com/google/adk-python/blob/1063fa532cad59d8e9f7421ce2f523724d49d541/src/google/adk/flows/llm_flows/contents.py#L127) only applies to the trailing part of the event chain, and will be still "compatible" with the processing later on, when non-response events arrive.

r4inee:
Thanks @tilgalas for the detailed review, that was what I was thinking too to align with the adk-python implementation. I think that will fix the error so I'll start with making that change.

Maybe we can treat this as a seperate problem, but the "superseeded" logic which you pointed out distorts the chronological sequences of messages (but not the events). `adk-python` also behaves this way but it seems like it can have an effect on the accuracy of subsequent model outputs.

Using a text example to make that clearer:

1. User asks "What is the weather in New York?"
2. FC - getWeather("New York")
3. FR1 - getWeather failed with error requiring confirmation
4. Model responds with "It seems like I need your confirmation, would you like me to proceed?"
5. FR2 - getWeather("New York") succeeds and returns 22 degrees
6. Model respond with "Weather in New York is 22 degrees"
7. User asks "What about weather in Sydney?"
...

After merging
1. User asks "What is the weather in New York?"
2. FC - getWeather("New York")
3. FR2 - getWeather("New York") succeeds and returns 22 degrees
4. Model respond with "It seems like I need your confirmation, would you like me to proceed?" <- **this message is out of order.**
6. Model respond with "Weather in New York is 22 degrees"
7. User asks "What about weather in Sydney?"
...

r4inee:
Hey @tilgalas - I have updated the implementation to match `adk-python` while we can have the discussion with the `adk-python` folks.

I tried to keep the change minimal since there are some bespoke `gemini-3` logic in that function.

r4inee:
Hi team! would like some help getting this merge and included in the next release (whenever available).

tilgalas:
Hi @r4inee thank you for your patience, I'll take a look at your changes as soon as possible!

tilgalas:
hi @r4inee apologies, but we had to revert the change as it turns out it breaks some of our internal tests. I'll try to figure out what exactly the issue was and get back to you - you could open another PR with the same change in the meantime, which we can collaborate on again if you'd like. I'll still try to come up with a fix for the bug if you'd rather not spend more time on this.

r4inee:
Hey @tilgalas - we can keep collaborating here while that happens to keep the conversations in one place, would also like to see if the adk-java team reaches a different implementation than adk-python for the aforementioned problem.

Happy to make a further contribution once we know what the issue was

tilgalas:
hey @r4inee, sure, sounds good - the good news is that I think I've found what the issue was on our side and if everything goes well we'll be able to bring back your fix, I'll keep you posted

tilgalas:
hi @r4inee, fyi we merged your change again in #832

r4inee:
Awesome! thanks for the help, and for my own learning, what was the the reason?

and did you also manage to discuss with the python folks internally on the other proposed change? or should I raise that seperately in `adk-python`?

tilgalas:
Oh, we had a test that mocked function responses that didn't have id field populated, with your change in place they were silently removed from the event history, the fix was to just add the ids

I think moving the discussions to adk-python would be a good idea, let me summarize your proposal though, just to see if I understand it correctly - you want to remove all the intermediate events between the function call and its response? That seems like too much to me - I imagine there could be some important events, unrelated to the FC-FR pair interleaved there. You mentioned the confirmation messages could be removed, so maybe I misrepresented your idea here - is there a robust way to determine which messages are "safe" to remove and which ones we must keep?

r4inee:
Yes exactly. 

Looking at this a bit closer, I think you are right that it may accidently remove some redundent event, and the redundant event currently doesn't have any attribute that I can reliably filter on, it looks just like any other model response.

```
Event.builder()
                        .author(AGENT_NAME)
                        .content(Content.builder()
                                .role("model")
                                .parts(List.of(Part.fromText("It seems like I need your confirmation, would you like me to proceed?")))
                                .build())
                        .build()
```

We would need an attribute on the Event class (similar to Event#partial) that indicate whether an event is filter-able.

PR #717, opened 2026-01-09
Fix A2A metadata key prefix for cross-language compatibility

fsonntag:
### Problem
When evaluating Java agents from Python via A2A protocol, tool calls were not being tracked because metadata keys didn't match Python's expectations.

**Root cause**: Python expects metadata keys with `adk_` prefix (e.g., `adk_type`), but Java was sending unprefixed keys (e.g., `type`).

### Solution
Added `ADK_METADATA_KEY_PREFIX = "adk_"` constant and prefixed metadata keys to align with Python's convention:
- `A2A_DATA_PART_METADATA_TYPE_KEY` → `"adk_type"`
- `A2A_DATA_PART_METADATA_IS_LONG_RUNNING_KEY` → `"adk_is_long_running"`

Metadata values (like `"function_call"`) remain unprefixed as expected.

### Changes
**File**: `a2a/src/main/java/com/google/adk/a2a/converters/PartConverter.java`
- Added `ADK_METADATA_KEY_PREFIX` constant at line 33
- Updated metadata key constants at lines 36-38 to use prefix
- Value constants (lines 39-43) correctly remain unprefixed

### Python Reference
Python's implementation using `_get_adk_metadata_key()` helper:
https://github.com/google/adk/blob/main/adk-python/src/google/adk/a2a/converters/part_converter.py#L36-L41

Python adds prefix when reading metadata:
https://github.com/google/adk/blob/main/adk-python/src/google/adk/a2a/converters/part_converter.py#L93-L98

### Testing
Verified with cross-language A2A evaluation showing tool calls now properly tracked.

### DIFF
--- a/core/src/main/java/com/google/adk/agents/BaseAgent.java
+++ b/core/src/main/java/com/google/adk/agents/BaseAgent.java
@@ -17,6 +17,7 @@
 package com.google.adk.agents;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static java.util.Arrays.stream;
 
 import com.google.adk.Telemetry;
 import com.google.adk.agents.Callbacks.AfterAgentCallback;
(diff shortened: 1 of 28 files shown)


anFatum:
Hello @fsonntag!
Sorry for late response, we've had a major refactoring of the A2A module, now the metadata keys were moved with prefixes to the https://github.com/google/adk-java/blob/main/a2a/src/main/java/com/google/adk/a2a/converters/A2AMetadataKey.java

I'm closing the PR, thank you!

PR #725, opened 2026-01-12
Remove ADK dependency for langchain4j module

gbrail:
The langchain4j module in the "contrib" directory depends on adk-dev, although it doesn't seem to need anything in that module. This, in turn, pulls in tons of stuff for applications that use it, including logback and Spring Boot, which I'd like to not have to include in my project. This PR just takes the dependency out.


vorburger (approved):
LGTM, but I'll let current maintainers of this repo do the actual merge.

glaforge (approved):

Issue #699, opened 2026-01-02
[spring-ai] NullPointerException in ToolConverter when ADK tool uses ToolContext


lurongjiang:
** Please make sure you read the contribution guide and file the issues in the rigth place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**
In `com.google.adk.models.springai.ToolConverter#convertToSpringAiTools`, when converting ADK BaseTool instances to Spring AI ToolCallback, the `toolContext` parameter passed to `tool.runAsync(...)` is hardcoded as null. This causes a NullPointerException if the underlying ADK tool method expects and uses a non-null ToolContext (e.g., to access functionCallId()).

**To Reproduce**
1. Define an ADK tool method that includes a `ToolContext` parameter.  
2. Use `ToolConverter` to convert this tool into a Spring AI `ToolCallback`.  
3. Invoke the resulting `ToolCallback.call(...)` with a valid `org.springframework.ai.chat.model.ToolContext`.  
4. Observe a `NullPointerException` because the ADK tool receives `null` for its `ToolContext` argument.

test case:
```java
@Test
void testCallbackWithToolContext() throws Exception {
  ToolConverter converter = new ToolConverter();
  FunctionTool tool = FunctionTool.create(TestSayHelloFunctions.class, "sayHello");
  Map<String, BaseTool> tools = Map.of("sayHello", tool);

  List<ToolCallback> toolCallbacks = converter.convertToSpringAiTools(tools);
  ToolCallback callback = toolCallbacks.get(0);
  String res = callback.call("{\"name\":\"Jack\"}", new org.springframework.ai.chat.model.ToolContext(new HashMap<>()));
  System.out.println(res);
}

public static class TestSayHelloFunctions {
  @Annotations.Schema(name = "sayHello", description = "Test func: say hello for name")
  public static void sayHello(
      @Annotations.Schema(name = "name", description = "the name to say hello") String name,
      @Annotations.Schema(name = "toolContext", description = "toolContext") ToolContext toolContext) {
    toolContext.functionCallId().ifPresent(functionCallId -> System.out.println("callId:" + functionCallId));
    System.out.println("hello " + name);
  }
}
```

**Expected behavior**
The ADK `ToolContext` passed to the tool method should be constructed from the `org.springframework.ai.chat.model.ToolContext` provided to `ToolCallback.call(...)`, rather than being `null`. At minimum, it should not cause an NPE when the tool expects a non-null context.

[...]

PR #733, opened 2026-01-15
Tool call fix for Spring AI - Let ADK call tools

alexcheng1982:
In the current implementation of Spring AI integration, tool calls are handled by Spring AI, this means the code needs to handle the conversion between Spring AI tools and ADK tools. This is unnecessary and introduces serval issues, like `ToolContext` is `null` (see #699 ), before and after tool callbacks are not invoked.

In this PR, tool calls are handled by ADK, just like other `BaseLlm` implementations.
- Spring AI tool calls execution is disabled by setting `internalToolExecutionEnabled` to `false` in `ToolCallingChatOptions`.
- `FunctionToolCallback` in Spring AI simply returns empty strings. `FunctionToolCallback` is only used to pass tool definitions. The actual tool calls are handled by ADK.
- Handles `ToolResponseMessage` in `MessageConverter`.
- Removed unused code and test.


glaforge:
cc @ddobrin for your info

ddobrin:
Hi @alexcheng1982 
There is the argument that tool calling could be handled, depending on the use-case, in either the LLMAgent or by SpringAI. 

It might be better that there is a (flexible) config which should allow for the selection of who should do function calling

ddobrin:
A couple of notes here:

1. Data loss in `handleUserContent()` when Content has both text and function response parts. 

The old code added both `UserMessage` and `toolResponseMessages` to the result list. 
The new code uses an if/else that returns *only* tool response messages when present, discarding any
 accumulated text and media content. 
This will cause data loss when a `Content` object contains both text parts and `functionResponse` parts.

 [link ](https://github.com/google/adk-java/blob/fba5f4fe41ac12f30ff68f9992051abbee17c2eb/contrib/spring-ai/src/main/java/com/google/adk/models/springai/MessageC
 onverter.java#L263-L269)

 2. Test coverage reduced with no replacement. 
 
  `ToolConverterArgumentProcessingTest.java` (128 lines, 4 test scenarios) was deleted along with the  `processArguments()` method. 
No new tests were added for the new behaviour: `ToolResponseMessage` creation from function responses (lines 209-228), the no-op tool function in `ToolConverter`, or the `internalToolExecutionEnabled(false)` configuration.

[link]( https://github.com/google/adk-java/blob/fba5f4fe41ac12f30ff68f9992051abbee17c2eb/contrib/spring-ai/src/main/java/com/google/adk/models/springai/MessageC
 onverter.java#L208-L229)

arthuursantos:
hi

the issue has been closed, but the PR remains open 

was this fixed?

ddobrin:
Could you please validate that the issue has been addressed at this time with the existing codebase?
There were 2 issues addressed with the same fix if I remember correctly
Thank you

arthuursantos:
hi again

the issue has not yet been resolved

i manually added the fixes from this PR and everything was normal, but when I removed them, the error persists

Caused by: java.lang.NullPointerException: Cannot invoke "com.google.adk.tools.ToolContext.eventActions()" because "toolContext" is null
at com.google.adk.flows.llmflows.AgentTransfer.transferToAgent(AgentTransfer.java:164) ~[google-adk-1.0.0.jar:na]
at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103) ~[na:na]
... 276 common frames omitted

dependencies:

```
<dependency>
<groupId>com.google.adk</groupId> 
<artifactId>google-adk</artifactId> 
<version>1.0.0</version> 
</dependency> 

<dependency> 
<groupId>com.google.adk</groupId> 
<artifactId>google-adk-spring-ai</artifactId> 
<version>1.0.0</version> 
</dependency>
```

glaforge:
@ddobrin Looking at the PR and the current source code, @arthuursantos is right and the changes in this PR don't seem to have been applied to the codebase right now.

hemasekhar-p:
Hi @alexcheng1982, To proceed with the review, could you please address the following,
2. Address the merge conflects
3. Resolve the current Maven build failures.

PR #888, opened 2026-02-22
Fix null Event ID in adk_request_confirmation events

### LINKED ISSUE #871: Missing Id generation for Event
**Describe the Bug:**

Events generated for tool confirmation requests (adk_request_confirmation) are being created with null IDs. This occurs when tools require user confirmation before execution. The root cause is that Functions.generateRequestConfirmationEvent() creates events without calling .id(Event.generateEventId()), resulting in events being persisted to session storage with id: null.

**Steps to Reproduce:**

1. Configure an agent with a tool that requires confirmation (e.g., using ConfirmationTool)
2. Run the agent and trigger the tool that requires confirmation
3. Inspect the generated event in the session
4. Observe that the event has "id": null

**Expected Behavior:**
All events should have a unique, non-null ID generated via Event.generateEventId().

**Observed Behavior:**
Event object is being created with "null" id 

**Environment Details:**

- ADK Library Version: 0.5.0 (current main branch)
- OS: All platforms (Java-based issue)
- Java Version: N/A (code issue)

**Model Information:**

 - Model agnostic
---

https://github.com/google/adk-java/blob/main/core/src/main/java/com/google/adk/flows/llmflows/Functions.java
method - generateRequestConfirmationEvent

```
return Optional.of(
        Event.builder()                                                             //missing id
            .invocationId(invocationContext.invocationId())
            .author(invocationContext.agent().name())
            .branch(invocationContext.branch())
            .content(contentBuilder.build())
            .longRunningToolIds(longRunningToolIds)
            .build());
```

Always

YuqiGuo105:
Fixes #871

Events generated for tool confirmation requests (adk_request_confirmation) were being created with null IDs because the generateRequestConfirmationEvent() method was missing the .id(Event.generateEventId()) call in the Event builder.

This commit:
- Adds .id(Event.generateEventId()) to generateRequestConfirmationEvent() in Functions.java to ensure all confirmation request events have non-null UUIDs
- Adds regression test generateRequestConfirmationEvent_generatesNonNullEventId() to FunctionsTest.java to verify the fix

### DIFF
--- a/core/src/main/java/com/google/adk/flows/llmflows/Functions.java
+++ b/core/src/main/java/com/google/adk/flows/llmflows/Functions.java
@@ -682,6 +682,7 @@ public final class Functions {
 
     return Optional.of(
         Event.builder()
+            .id(Event.generateEventId())
             .invocationId(invocationContext.invocationId())
             .author(invocationContext.agent().name())
             .branch(invocationContext.branch().orElse(null))
--- a/core/src/test/java/com/google/adk/flows/llmflows/FunctionsTest.java
+++ b/core/src/test/java/com/google/adk/flows/llmflows/FunctionsTest.java
@@ -26,6 +26,8 @@ import com.google.adk.agents.InvocationContext;
 import com.google.adk.agents.RunConfig;
 import com.google.adk.agents.RunConfig.ToolExecutionMode;
 import com.google.adk.events.Event;
+import com.google.adk.events.EventActions;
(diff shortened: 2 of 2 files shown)


yongkiy-google:
Hi, thank you for the contribution. We’ve addressed this in https://github.com/google/adk-java/pull/1092 instead to ensure the behavior stays close to the Python ADK. We really appreciate you taking the time to help out!

PR #942, opened 2026-03-03
dev: Introduce initial AGENTS.md

vorburger:
Intentionally named AGENTS.md instead of e.g. GEMINI.md to be fully model neutral; see https://agents.md for background.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,3 @@
+# AGENTS.md
+
+Validate changes by running `./mvnw test`.


AGENTS.md
```diff
@@ -0,0 +1,3 @@
+# AGENTS.md
+
+Validate changes by running `./mvnw install`.
```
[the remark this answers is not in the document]
vorburger: Nah, I think this is clear enough for coding agents.

krwc: Not mvnw test or similar?

vorburger: > Not mvnw test or similar?

Actually you are right, `test` would be better than `install` (so that Agents detect regressions). 

Let me change that - give me 2 minutes.

vorburger: @krwc done - PTAL?

krwc: Great, thanks!

PR #956, opened 2026-03-04
Extract timestamp as double for InMemorySessionService events

tlanclos:
InMemorySessionService sets a Session's last modified time based on the last appended event's timestamp. The timestamp in an event is recorded in millis while the Session's timestamp is an Instant. During the transformation, Events perform this converstion using division. Before this change, the timestamp was truncated to the second, yet the code was trying to extract nanos which were always 0. This fixes that bug with a simple type change. I've also added a test to prevent regressions.

### DIFF
--- a/core/src/main/java/com/google/adk/sessions/InMemorySessionService.java
+++ b/core/src/main/java/com/google/adk/sessions/InMemorySessionService.java
@@ -154,17 +154,11 @@ public final class InMemorySessionService implements BaseSessionService {
     if (config.numRecentEvents().isEmpty() && config.afterTimestamp().isPresent()) {
       Instant threshold = config.afterTimestamp().get();
 
-      eventsInCopy.removeIf(
-          event -> getEventTimestampEpochSeconds(event) < threshold.getEpochSecond());
+      eventsInCopy.removeIf(event -> getInstantFromEvent(event).isBefore(threshold));
     }
 
     // Merge state into the potentially filtered copy and return
     return Maybe.just(mergeWithGlobalState(appName, userId, sessionCopy));
-  }
-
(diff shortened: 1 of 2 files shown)


anFatum:
Hello @tlanclos! Thank you for your contribution!

I think we actually could change the `getInstantFromEvent` to the 
```
  private Instant getInstantFromEvent(Event event) {
    return Instant.ofEpochMilli(event.timestamp());
  }
```
since event.timestamp() returns milis.

I'm quite unsure about changing the type from long to double, given that we have a comparison of double and long `getEventTimestampEpochSeconds(event) < threshold.getEpochSecond()`, where threshold.getEpochSecond() returns long

I've checked and your test pass with the changed getInstantFromEvent method implementation. What do you think?

Additionally, I guess we might get rid of the `getEventTimestampEpochSecond` and achieve a cleaner result by replacing it with 
```
event -> event.timestamp() < threshold.toEpochMilli()
```

tlanclos:
Thanks for the suggestion, @anFatum! I originally took the approach that I did so that I introduced minimal code change; however, I agree that the type handling could be better. Instead of comparing against longs or doubles, I decided to remove `getEventTimestampEpochSeconds` and instead move the conversion to `getInstantFromEvent` since that method is used in a few places. I think the next step in cleaning this up would be to make `Event.timestamp()` an `Instant` as suggested in the TODOs.

anFatum:
Thanks for your change @tlanclos! Looks good to me, the only thing I'll ask you to squash your commits so I can merge this PR, thank you!

tlanclos:
No problem, got everything squashed! It should be ready for merge

PR #1072, opened 2026-03-23
fix: escape sessionId in VertexAiClient request paths

Tulgaaaaaaaa:
### Problem

`VertexAiClient` concatenates the session id straight into the Vertex AI REST path:

```java
"reasoningEngines/" + reasoningEngineId + "/sessions/" + sessionId
```

Session ids reach the client from user-supplied `Session` objects, so a session id
containing `/` or `..` can retarget the request at a different resource, and one
containing `?` can append arbitrary query parameters to the request.

### Fix

Escape the session id as a single URL path segment with
`UrlEscapers.urlPathSegmentEscaper()` in `listEvents`, `getSession`, `deleteSession`
and `appendEvent`.

Two values in this class were already safe and are left alone:

- `reasoningEngineId` is constrained to digits by `VertexAiSessionService.parseReasoningEngineId`.
- `userId` in `listSessions` is already wrapped as a quoted AIP-160 literal and
  form-escaped on `main`, so the original version of this PR (which replaced that
  with a bare `URLEncoder.encode`) would have been a regression. Rebased away.

`UrlEscapers` is used rather than `URLEncoder.encode`, which is HTML form encoding —
it turns a space into `+`, which is a literal `+` in a path segment.

### Tests

New `VertexAiClientTest` — 12 tests using Mockito `ArgumentCaptor` to assert on the
exact path handed to `HttpApiClient.request`:

- path traversal is neutralised: `../../secret` → `..%2F..%2Fsecret`
- query-string injection is neutralised: `456?view=FULL` → `456%3Fview=FULL`
- normal ids are untouched: `reasoningEngines/123/sessions/456`
- the `:appendEvent` custom verb and the `listEvents` `filter` parameter still survive escaping
- `listSessions` regression cover for the existing AIP-160 quoting

### Notes for reviewers

Rebased onto current `main` as a single commit — merge conflicts resolved, and the
`listEvents(reasoningEngineId, sessionId, filter)` signature that the earlier version
of this branch had dropped is preserved, which is what was breaking the Maven build.

Verified locally: `./mvnw test -Dtest=VertexAiClientTest` (12/12), the existing
`VertexAiSessionServiceTest` (30/30), and a full-project `test-compile`.


hemasekhar-p:
Hi @Tulgaaaaaaaa, thank you for your contribution! We appreciate you taking the time to submit this pull request. To proceed with the review, could you please address the following,
1. Please address the merge conflicts
2. As per our contribution policy, please squash your changes into a single commit 
3. Please resolve the current Maven build failures.

Tulgaaaaaaaa:
Hi, I've addressed the requested changes:

- Resolved the merge conflicts
- Squashed the commits into a single commit
- Fixed the Maven build issues

Could you please review the updated PR when you have time? Thank you!

hemasekhar-p:
@Tulgaaaaaaaa, thank you for addressing the feedback and making those updates. Currently this PR is under review by our team, we will keep you posted if any additional information is required. thank you

sherryfox:
@Tulgaaaaaaaa, thank you for your contribution, and we appreciate you taking the time to submit this pull request. The exact vulnerability described in the PR body was already closed on main on 2026-07-03 by #1323 (commit d1b1d927f).

That change added `validateSessionId` to `VertexAiSessionService` (`^[a-zA-Z0-9_-]+$`) and applies it on every path reaching `VertexAiClient`, which is package-private and used only by that service — so a session id containing `/`, `?` or `%` is rejected before it can reach a request path, and `urlPathSegmentEscaper` would be a no-op on whatever survives.

The vulnerability was real when you filed this in March; it was closed by a different change rather than because anything here was wrong. Sorry it sat open as long as it did. Closing as obsolete.

PR #1073, opened 2026-03-23
fix: parallel agent execution

glaforge:
Follow up to #610 as a test was failing:
```
ParallelAgentTest.runAsync_subAgentsExecuteInParallel_eventsOrderedByCompletion:101 » NoSuchElement No value present
```
The test was failing because the branch was empty. This happened because the `BaseAgent` would only "append" its name to an existing branch. If the root agent was started with no branch (which is common in tests), its context—and subsequently its sub-agents' contexts—remained branchless.

### DIFF
--- a/core/src/test/java/com/google/adk/agents/ParallelAgentTest.java
+++ b/core/src/test/java/com/google/adk/agents/ParallelAgentTest.java
@@ -25,7 +25,10 @@ import com.google.common.collect.ImmutableList;
 import com.google.genai.types.Content;
 import com.google.genai.types.Part;
 import io.reactivex.rxjava3.core.Flowable;
+import io.reactivex.rxjava3.core.Scheduler;
 import io.reactivex.rxjava3.schedulers.Schedulers;
+import io.reactivex.rxjava3.schedulers.TestScheduler;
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/google/adk/agents/ParallelAgentTest.java
```diff
@@ -110,4 +119,79 @@ public void runAsync_noSubAgents_returnsEmptyFlowable() {
...
+        ParallelAgent.builder().name("parallel_agent").subAgents(agent1, agent2).build();
+
+    InvocationContext invocationContext = createInvocationContext(parallelAgent);
+
+    long startTime = System.currentTimeMillis();
+    List<Event> events = parallelAgent.runAsync(invocationContext).toList().blockingGet();
+    long duration = System.currentTimeMillis() - startTime;
+
+    assertThat(events).hasSize(2);
+    // If parallel, duration should be less than 1.5 * sleepTime (1500ms).
```
krwc, line 167: I think it'd be better to use locks / cvars to test this.

I won't block the submission on it, but relying on sleeps tends to lead to test flakiness.

krwc:
We will resolve this in https://github.com/google/adk-java/pull/1077.

PR #1105, opened 2026-04-01
feat: Binding url citation metadata for event

flex-myeonghyeon:
## Summary

- Bind URL citation metadata to events when processing model responses
- Extract and attach citation source URLs from grounding metadata to the corresponding event
- Enables downstream consumers to access citation information directly from events

## Test plan

- [ ] Verify that citation metadata is correctly bound to events when URL citations are present
- [ ] Verify that events without citation metadata are unaffected
- [ ] Run existing unit tests to confirm no regressions

<details>
<summary>이 작업을 진행하는데 사용된 프롬프트</summary>

```
https://github.com/flex-myeonghyeon/adk-java 로 dc81846cc58c367dc72dab0f0eb264a8e5b4cdaf 체리픽 하고 PR 올려줘
```

</details>


glaforge:
As a point of reference: in Python, the citation metadata is kept at the level of the equivalent `LlmResponse` object.
https://github.com/google/adk-python/blob/bbad9ec64ce1617bc45148de97e6246752845b98/src/google/adk/models/llm_response.py#L132

Wouldn't that be a better place for holding the citation metadata information?
Wouldn't that suit your need to have access to this metadata?

flex-myeonghyeon:
@glaforge 
That is correct. First of all, I have added the citationMetadata property to the LlmResponse class.

Regarding your question, I believe you were asking about adding the property to the Event class. Since the Event class already contains groundingMetadata, I thought citationMetadata belonged at the same level of information and added it there as well:
https://github.com/google/adk-java/blob/d348a2a59710948c9c1fd8e2cf7fe98fae2df0d0/core/src/main/java/com/google/adk/events/Event.java#L64

Do you think it should only be added to LlmResponse? If so, I will remove it from the Event class. Please let me know your thoughts.

glaforge:
@flex-myeonghyeon Hmmm... That's interesting...
I was pointing at Python to see what they had done there, but their `event` derives from `llm response` which has that citation metadata. But in Java, our `Event` class doesn't extend `LlmResponse` at all.
And indeed, I also noticed that `GroundingMetadata` was on `Event`, so `CitationMetadata` could very well have its place there.

I'm also curious: what lead you to this pull request? When/how did you need the citation metadata? What was the use case?

flex-myeonghyeon:
@glaforge 
I am currently using the GoogleSearchAgentTool.
https://github.com/google/adk-java/blob/d348a2a59710948c9c1fd8e2cf7fe98fae2df0d0/core/src/main/java/com/google/adk/tools/GoogleSearchAgentTool.java#L29

When the GoogleSearchAgentTool is used in an ADK response, I wanted to be able to verify the search information that was referenced. My primary goal is to validate the basis of the response, but I also intend to pass this as citation information to the client if needed.

While I noticed that the Event class already provides groundingMetadata, I realized that it was missing citationMetadata. That is why I decided to add it.

flex-myeonghyeon:
@glaforge Any update?

hemasekhar-p:
Hi @flex-myeonghyeon, thank you for your contribution. As per contribution policy, could you please squash your commits into a single commit message?  and the Maven build is currently failing with your changes. could you please investigate and resolve these build issues?

hemasekhar-p:
This PR is being closed due to a lack of updates over the past couple of weeks. Feel free to reopen it as soon as the review notes have been addressed.

PR #1137, opened 2026-04-14
fix: Make LlmResponse constructor protected for external subclassing

### LINKED ISSUE #1134: LlmResponse constructor makes it hard to create new response type outside of package
I am trying to connect LM-Studio as a local AI from the ADK. For that, I have to implement some specific classes, like LMStudioRequest (extends LlmRequest) etc.

But LMStudioResponse is hard to implement, because I cannot overwrite the LlmResponse constructor. What am I doing wrong? Or could the empty constructor be removed from the LlmResponse class?

Thanks in advance

jinnigu:

- Closes: #1134

**Problem:**
The `LlmResponse` constructor uses package-private (default) visibility, which prevents users from subclassing `LlmResponse` outside of the `com.google.adk.models` package. This blocks custom LLM integrations (e.g., connecting LM Studio or other local models) that need to extend `LlmResponse`.

**Solution:**
Change the `LlmResponse` constructor from package-private to `protected`. This is the minimal change needed. It allows subclassing from external packages while still preventing arbitrary instantiation (unlike `public`). The `@AutoValue`-generated subclass in the same package is unaffected.

### Testing Plan

**Unit Tests:**

Added `ExternalLlmResponseSubclassTest` in the `com.google.adk.models.external` package (outside `com.google.adk.models`) to verify that an external subclass of `LlmResponse` can be instantiated.

```
./mvnw -pl core -Dtest=ExternalLlmResponseSubclassTest test
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 — BUILD SUCCESS
```

### Checklist

### DIFF
--- a/core/src/main/java/com/google/adk/models/LlmResponse.java
+++ b/core/src/main/java/com/google/adk/models/LlmResponse.java
@@ -41,7 +41,8 @@ import org.jspecify.annotations.Nullable;
 @JsonDeserialize(builder = LlmResponse.Builder.class)
 public abstract class LlmResponse extends JsonBaseModel {
 
-  LlmResponse() {}
+  /** Protected to allow custom LLM integrations to subclass this type outside the package. */
+  protected LlmResponse() {}
 
   /**
    * Returns the content of the first candidate in the response, if available.
--- a/core/src/test/java/com/google/adk/models/external/ExternalLlmResponseSubclassTest.java
+++ b/core/src/test/java/com/google/adk/models/external/ExternalLlmResponseSubclassTest.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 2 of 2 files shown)


damianmomotgoogle:
Hi!

Could you explain more on why you aim to subclass this type ?

I don't think we should make LlmResponse's constructor protected — subclassing it isn't the intended extension point.

`LlmResponse` is an `@AutoValue` immutable value type. The package-private constructor is deliberate: the only subclass that's meant to exist is the generated `AutoValue_LlmResponse`. 

For the underlying use case (#1134, integrating LM Studio / a custom model), the right extension point is `BaseLlm`, not the value types. `BaseLlm.generateContent(...)` returns `Flowable<LlmResponse>` — you build responses with `LlmResponse.builder()...build()` rather than subclassing. That's how all existing non-Gemini integrations do it. If a required field is missing on `LlmResponse`, the fix is to add that field (or use `customMetadata()`), not to open up subclassing.

hemasekhar-p:
I am closing this PR for now, as it has been stale for more than two weeks. Please feel free to reopen it once the review comments have been addressed.

PR #1160, opened 2026-04-26
Multi-agent patterns examples

svetanis:
This project demonstrates various multi-agent patterns using the ADK (Agent Development Kit) framework. It showcases how multiple AI agents can collaborate to perform complex tasks across different domains, including code generation, report writing, and travel planning.

----
Title: Fix NullPointerException in AgentTool when Agent description is missing

The Problem
When an Agent is defined without a description, calling AgentTool.declaration() fails with a NullPointerException.

The crash occurs because the adk-java library passes a null description to the google-genai SDK's FunctionDeclaration.Builder. The generated builder implementation for FunctionDeclaration uses Optional.of(description), which is not null-safe.

Stack Trace snippet:

java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:220)
	at java.base/java.util.Optional.of(Optional.java:113)
	at com.google.genai.types.AutoValue_FunctionDeclaration$Builder.description(AutoValue_FunctionDeclaration.java:160)
	at com.google.adk.tools.AgentTool.declaration(AgentTool.java:135)
	
Suggested Fix (Upstream)
Instead of the default generated setter, the builder should use Optional.ofNullable()

Current generated logic:

public FunctionDeclaration.Builder description(String description) {
  this.description = Optional.of(description); 
  return this;
}

Proposed manual override in the abstract Builder class:

 public Builder description(String description) {
        return description(Optional.ofNullable(description));
    }
    // ... apply same pattern to other Optional fields

Temporary Workaround for adk-java
Until the upstream google-genai library is updated, adk-java could protect itself in AgentTool.java:135 by ensuring it never passes a null or empty string to the builder:

  @Override

    // temporary workaround for adk-java to protect itself
    // to ensure it never passes a null or empty string
    // to the FunctionDeclaration builder
    // until the upstream google-genai library is updated
    String desc = isNotBlank(this.description()) ? this.description() : this.name();

    FunctionDeclaration.Builder builder =
        FunctionDeclaration.builder().description(desc).name(this.name());

### DIFF
--- a/contrib/multi-agent-patterns/.gitignore
+++ b/contrib/multi-agent-patterns/.gitignore
@@ -1,0 +1,29 @@
+# Compiled class file
+*.class
+*.classpath
+*.project
+*.settings
+*.factorypath
+target/
+
+# Log file
+*.log
+
+# BlueJ files
+*.ctxt
+
+# Mobile Tools for Java (J2ME)
+.mtj.tmp/
+
+# Package Files #
+*.jar
+*.war
+*.nar
+*.ear
+*.zip
+*.tar.gz
+*.rar
+
+# virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
+hs_err_pid*
+replay_pid*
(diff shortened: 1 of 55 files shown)


svetanis:
Closing this PR to split the bug fix and the multi-agent patterns into separate, cleaner PRs

PR #1164, opened 2026-04-28
fix(core): avoid NPE when Agent description is missing

svetanis:
Prevents a NullPointerException in the google-genai library when an Agent is defined without a description. Includes a regression test.

- Closes: _N/A_
- Related: _N/A_

_If applicable, please follow the issue templates to provide as much detail as
possible._

**Problem:**
_Calling AgentTool.declaration() triggers a NullPointerException when an Agent is defined without an explicit description. The crash occurs because the underlying google-genai library's generated FunctionDeclaration builder uses Optional.of() for the description field, which is not null-safe._

**Solution:**
_Coerce a null description to an empty string in AgentTool.declaration() using Guava's Strings.nullToEmpty(), so the google-genai builder always receives a non-null value. This matches the Python, Go, and Kotlin ports, which send an empty description when an agent has none._

### Testing Plan

_Please describe the tests that you ran to verify your changes. This is required
for all PRs that are not small documentation or typo fixes._

**Unit Tests:**

**Summary of passed java test results:**
_Passed AgentToolTest.declaration_withNullDescription_usesEmptyDescription, which verifies that a null description no longer triggers an NPE and results in an empty description._

_Verified by running a sample agentic workflow where the Agent class lacked a description field. Previously, this resulted in a crash during tool registration; it now initializes and executes correctly._

### Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas.

### Additional context
Stack Trace:
java.lang.NullPointerException
at java.base/java.util.Objects.requireNonNull(Objects.java:220)
at java.base/java.util.Optional.of(Optional.java:113)
at com.google.genai.types.AutoValue_FunctionDeclaration$Builder.description(AutoValue_FunctionDeclaration.java:160)
at com.google.adk.tools.AgentTool.declaration(AgentTool.java:135)

_Note on Root Cause:
While this PR ensures stability within adk-java, the underlying issue appears to be in the google-genai library. The generated AutoValue_FunctionDeclaration.Builder uses Optional.of() for the description field, which enforces a non-null constraint. A more robust upstream fix would involve the generator using Optional.ofNullable(). This local fix is necessary to prevent crashes in the meantime._

### DIFF
--- a/core/src/main/java/com/google/adk/tools/AgentTool.java
+++ b/core/src/main/java/com/google/adk/tools/AgentTool.java
@@ -31,6 +31,7 @@ import com.google.adk.runner.InMemoryRunner;
 import com.google.adk.runner.Runner;
 import com.google.adk.sessions.State;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.genai.types.Content;
@@ -131,8 +132,14 @@ public class AgentTool extends BaseTool {
 
   @Override
   public Optional<FunctionDeclaration> declaration() {
+
+    // The genai FunctionDeclaration builder uses Optional.of() internally, which NPEs on a null
+    // description. Coerce null to an empty string. This matches the Python, Go, and Kotlin ports,
+    // which send an empty description when the agent has none.
+    String desc = Strings.nullToEmpty(this.description());
+
     FunctionDeclaration.Builder builder =
-        FunctionDeclaration.builder().description(this.description()).name(this.name());
+        FunctionDeclaration.builder().description(desc).name(this.name());
 
     Optional<Schema> agentInputSchema = getInputSchema(agent);
 
--- a/core/src/test/java/com/google/adk/tools/AgentToolTest.java
+++ b/core/src/test/java/com/google/adk/tools/AgentToolTest.java
@@ -43,6 +43,7 @@ import com.google.genai.types.Schema;
 import io.reactivex.rxjava3.core.Flowable;
 import io.reactivex.rxjava3.core.Maybe;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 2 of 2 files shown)


hemasekhar-p:

svetanis:
@hemasekhar-p ,

Thanks for the update! 
Just a quick heads up - I went ahead and rebased the branch onto the latest `main`.

core/src/main/java/com/google/adk/tools/AgentTool.java
```diff
@@ -16,6 +16,8 @@
 
 package com.google.adk.tools;
 
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
```
kvmilos: can we use Guava's `Strings.nullToEmpty` or JDK's `String.isBlank()` to not add a new dependency?

core/src/main/java/com/google/adk/tools/AgentTool.java
```diff
@@ -131,8 +133,11 @@ private Optional<Schema> getOutputSchema(BaseAgent agent) {
 
   @Override
   public Optional<FunctionDeclaration> declaration() {
+
+    String desc = isNotBlank(this.description()) ? this.description() : this.name();
```
kvmilos: One thing to align on: the Python, Go, and Kotlin ports all send an empty description when an agent doesn't have one — none of them use the name. Since this field is what the model reads to decide whether to call the tool, the name is a bit misleading there. Can we match the other ports

core/src/test/java/com/google/adk/tools/AgentToolTest.java
```diff
@@ -882,4 +883,20 @@ private ToolContext createToolContext(BaseAgent agent) {
...
+    // Create an agent with a name but NO description
+    LlmAgent testAgent =
+        LlmAgent.builder()
+            .name("TestAgent")
+            // .description() is intentionally left out
+            .build();
+
+    AgentTool tool = AgentTool.create(testAgent);
+    Optional<FunctionDeclaration> declaration = tool.declaration();
+    assertThat(declaration.isPresent());
```
kvmilos: I believe that should be`assertThat(declaration).isPresent();`

kvmilos (changes requested):
Thanks @svetanis for catching the bug and for the PR. Just a few small things below.

svetanis:
Thanks for the review, @kvmilos! All three points addressed:

1. **Dependency** — dropped `commons-lang3.StringUtils.isNotBlank` for Guava's `Strings.nullToEmpty`.

2. **Port parity** — now sends an empty description when the agent has none, matching the Python, Go, and Kotlin ports. 
This also simplified the fix to a single line with no name fallback:
   ```java
   String desc = Strings.nullToEmpty(this.description());
   ```

4. **Test** — fixed the no-op assertion (`assertThat(declaration).isPresent();`), renamed the test to `declaration_withNullDescription_usesEmptyDescription`, and asserted the empty description via `hasValue("")`.

All 27 `AgentToolTest` tests pass. 
The branch has also been rebased against the latest `main`.

kvmilos (approved):

Shimulmk (commented):
9vd6P1chE1vE6L7scGabHZc5jexGabARN5DKtYKt8HAi

PR #1183, opened 2026-05-05
fix: resolve missing trace spans in embedded AdkWebServer

### LINKED ISSUE #1175: Web UI "Trace" tab is empty because Dev UI's `OpenTelemetrySdk` is never bridged into `com.google.adk.telemetry.Tracing`
**Describe the Bug:**

When running the embedded ADK Dev UI (`com.google.adk.web.AdkWebServer`) inside a Spring Boot host service, the **Trace tab is always empty**. Requests succeed, chat reaches the agent, tools execute, but no agent/tool spans are recorded against the session.

Root cause is a wiring gap in `google-adk-dev`: `OpenTelemetryConfig` builds an `OpenTelemetrySdk` carrying its `ApiServerSpanExporter` (the bean the Trace REST controllers read from), but builds it as a **non-global instance** and **never installs it into `com.google.adk.telemetry.Tracing`**. ADK runtime tracing is hard-bound to `GlobalOpenTelemetry.getTracer("gcp.vertex.agent")`, so it emits to the JVM-wide no-op default and nothing reaches the Dev UI exporter.

Two problems compound:

1. **Misleading log line / dead branch.** `GlobalOpenTelemetry.get()` does not throw when nothing is registered, it returns a no-op default. So the `try` block in `openTelemetrySdk()` is always entered, and a non-global SDK is always built. The `IllegalStateException` branch (which calls `buildAndRegisterGlobal()`) is effectively dead code under normal usage. The log line `"OpenTelemetry already registered globally, creating non-global instance."` is actively misleading: it fires even when nothing is globally registered.

2. **No `Tracing.setTracerForTesting(...)` call.** Even if the Dev UI's local SDK were built correctly, `Tracing.getTracer()` continues to read from `GlobalOpenTelemetry`. The Dev UI's `ApiServerSpanExporter` therefore sees zero spans from agent invocations.

**Steps to Reproduce:**

1. Embed the Dev UI in a Spring Boot service via the standard recipe (child Spring context launched from `AdkWebServer`).
2. Disable any external OpenTelemetry registration (no Langfuse, no Java agent, no actuator OTel autoconfig).
3. Boot the service, open the Dev UI on its configured port, and send a message to any agent.
4. Open the **Trace** tab in the left panel.

**Expected Behavior:**

The Trace tab displays an agent/tool span tree for the completed invocation.

**Observed Behavior:**

The Trace tab is empty. No spans are recorded. With `--logging.level.com.google.adk.web.config=DEBUG` enabled, the following is logged:

```
c.g.a.w.c.OpenTelemetryConfig — Configuring SdkTracerProvider with ApiServerSpanExporter.
c.g.a.w.c.OpenTelemetryConfig — Configuring OpenTelemetrySdk and registering globally.
c.g.a.w.c.OpenTelemetryConfig — OpenTelemetry already registered globally, creating non-global instance.
```

The "already registered globally" branch is taken even when nothing has registered globally.

**Environment Details:**

- ADK Library Version: `com.google.adk:google-adk:1.2.0`, `com.google.adk:google-adk-dev:1.2.0`
- OpenTelemetry SDK: `1.55.0`
- Java: 21
- Spring Boot: 4.0.5
- OS: N/A (reproduced across environments)

**Model Information:** N/A (issue is independent of model)

---

## 🟡 Optional Information

**Regression:** Unknown — N/A to our setup.

**Logs:**

```text
c.g.a.w.c.OpenTelemetryConfig — Configuring SdkTracerProvider with ApiServerSpanExporter.
c.g.a.w.c.OpenTelemetryConfig — Configuring OpenTelemetrySdk and registering globally.
c.g.a.w.c.OpenTelemetryConfig — OpenTelemetry already registered globally, creating non-global instance.
```

**Additional Context:**

Relevant decompiled sources from the affected versions:

`com.google.adk.telemetry.Tracing#getTracer()` (`google-adk:1.2.0`):

```java
public static Tracer getTracer() {
    if (tracer == null) {
        tracer = GlobalOpenTelemetry.getTracer("gcp.vertex.agent");
    }
    return tracer;
}
```

`com.google.adk.web.config.OpenTelemetryConfig#openTelemetrySdk(...)` (`google-adk-dev:1.2.0`):

```java
@Bean
public OpenTelemetry openTelemetrySdk(SdkTracerProvider sdkTracerProvider) {
    log.debug("Configuring OpenTelemetrySdk and registering globally.");
    try {
        GlobalOpenTelemetry.get(); // does NOT throw on no-op default — always enters this branch
        log.debug("OpenTelemetry already registered globally, creating non-global instance.");
        return OpenTelemetrySdk.builder().setTracerProvider(sdkTracerProvider).build(); // non-global
    } catch (IllegalStateException e) {
        log.debug("Registering OpenTelemetry globally.");
        OpenTelemetrySdk sdk = OpenTelemetrySdk.builder()
                .setTracerProvider(sdkTracerProvider)
                .buildAndRegisterGlobal();
        Runtime.getRuntime().addShutdownHook(new Thread(sdk::close));
        return sdk;
    }
}
```

**Impact:**

- The Trace tab is silently broken in the most common dev setup (Dev UI as the only OTel consumer). There is no warning log pointing at the wiring gap.
- The Trace tab is also broken when the host registers its own global SDK (e.g. Langfuse, GCP, Datadog), because the Dev UI's local non-global SDK receives no events from `GlobalOpenTelemetry`.
- Affects all supported deployment shapes: standalone `AdkWebServer`, embedded Dev UI in a host Spring Boot context, and the `runAdkDevUi` Gradle task.

**How often has this issue occurred?**

- ✅ Always (100%)

hemasekhar-p:

- Closes: #1175

**Problem:**
When `AdkWebServer` is embedded as a child Spring Context, `OpenTelemetry` spans generated by the core Tracing utility is dropped, leaving the `Dev UI's` Trace tab completely empty. This occurs because the `Dev UI` correctly skips `GlobalOpenTelemetry` registration to avoid polluting the parent application's telemetry, but the core engine's `Tracing` utility is hardcoded to only pull from the global registry, resulting in a fallback No-Op tracer.

**Solution:**
**Core:** Added a thread-safe `Tracing.setTracer()` method in `google-adk` to allow external frameworks to inject a custom tracer.
 **Web:** Updated `OpenTelemetryConfig` in `google-adk-dev` to automatically inject its isolated SDK tracer into the core engine upon bean creation.

### Testing Plan

1. Created a Spring Boot application EnterpriseAgentApplication configured as the parent context, launching `AdkWebServer` as a child context on port 8080.

2. Disabled external `OpenTelemetry` exports in application.properties (management.otlp.tracing.export.enabled=false) to ensure strict isolation.

3. Sent multiple prompts to the compiled agent.

4. Opened the `Trace` tab in the left-hand navigation panel.

### DIFF
--- a/dev/src/main/java/com/google/adk/web/config/OpenTelemetryConfig.java
+++ b/dev/src/main/java/com/google/adk/web/config/OpenTelemetryConfig.java
@@ -60,18 +60,14 @@ public class OpenTelemetryConfig {
 
     // Check if OpenTelemetry has already been set globally (common in tests)
     try {
-      io.opentelemetry.api.GlobalOpenTelemetry.get();
-      // If we get here, it's already set, so just return a new instance without global
-      // registration
-      otelLog.debug("OpenTelemetry already registered globally, creating non-global instance.");
-      return OpenTelemetrySdk.builder().setTracerProvider(sdkTracerProvider).build();
-    } catch (IllegalStateException e) {
-      // GlobalOpenTelemetry hasn't been set yet, safe to register globally
       otelLog.debug("Registering OpenTelemetry globally.");
-      OpenTelemetrySdk otelSdk =
+      OpenTelemetrySdk sdk =
           OpenTelemetrySdk.builder().setTracerProvider(sdkTracerProvider).buildAndRegisterGlobal();
-      Runtime.getRuntime().addShutdownHook(new Thread(otelSdk::close));
-      return otelSdk;
+      Runtime.getRuntime().addShutdownHook(new Thread(sdk::close));
+      return sdk;
+    } catch (IllegalStateException e) {
+      otelLog.debug("OpenTelemetry already registered globally, returning non-global instance.");
+      return OpenTelemetrySdk.builder().setTracerProvider(sdkTracerProvider).build();
     }
   }
 }


hemasekhar-p:
@krwc Please have a look into it.

dev/src/main/java/com/google/adk/web/config/OpenTelemetryConfig.java
```diff
@@ -64,13 +65,17 @@ public OpenTelemetry openTelemetrySdk(SdkTracerProvider sdkTracerProvider) {
       // If we get here, it's already set, so just return a new instance without global
       // registration
       otelLog.debug("OpenTelemetry already registered globally, creating non-global instance.");
+      OpenTelemetrySdk sdk =
+          OpenTelemetrySdk.builder().setTracerProvider(sdkTracerProvider).build();
+      Tracing.setTracer(sdk.getTracer("gcp.vertex.agent"));
       return OpenTelemetrySdk.builder().setTracerProvider(sdkTracerProvider).build();
```
sylwestergryzio: an observation: the builder method is called here second time. I suppose the sdk can be returned.

damianmomotgoogle: +1

damianmomotgoogle:
I traced how same is achieved in Python ADK and Go ADK

 1. Go ADK Approach
  The Go implementation uses a Providers struct that acts as a configuration helper.
   * Mechanism: It provides a SetGlobalOtelProviders() method.
   * Behavior: This method calls the standard OTel otel.SetTracerProvider() and logglobal.SetLoggerProvider().
   * Philosophy: It doesn't store a tracer instance in a private variable within the library. Instead, it configures the global OpenTelemetry state, and the library then retrieves its tracer from that global state (using
     otel.GetTracerProvider().Tracer(...)).

  2. Python ADK Approach
  The Python implementation is even more integrated with the OpenTelemetry SDK.
   * Mechanism: It uses a maybe_set_otel_providers() function in telemetry/setup.py.
   * Behavior: This function calls trace.set_tracer_provider(). If a provider was already set globally by the user, the ADK respects it and avoids overriding.
   * Philosophy: Like Go, it does not provide a library-specific set_tracer method. It assumes that if you want to change the tracer, you should configure the global OpenTelemetry environment.

  3. Comparison with the Java PR (1183)
  Your proposed public static void setTracer(Tracer) in Java adds a library-specific static setter. While this is simple, it differs from the other languages in a few key ways:

   * Bypasses Global Registry: In Java, the standard way to configure tracing is GlobalOpenTelemetry.set(OpenTelemetry). Providing Tracing.setTracer(Tracer) creates a secondary path that only affects the ADK.
   * Potential Inconsistency: If a user sets a custom Tracer via your new method but also configures GlobalOpenTelemetry for the rest of their app, they might end up with fragmented traces or inconsistent configuration.
   * Inversion of Control: Python and Go lean into the "Global Registry" pattern. The library says, "I will use whatever is in the global registry," and provides a helper to put things there. Your Java proposal says, "I
     have my own slot for a tracer, please give it to me."

  Recommendation
  To align with the Python and Go ADK architectures and follow idiomatic OpenTelemetry Java patterns, you might consider:

   1. Rely on GlobalOpenTelemetry: The current implementation already defaults to GlobalOpenTelemetry.getTracer("gcp.vertex.agent"). Since this returns a lazy wrapper, users can simply call GlobalOpenTelemetry.set(...) at
      application startup, and the ADK will automatically pick up the custom configuration.




Would it work if we set GlobalOpenTelemetry.set(...)  in ADK Web instead ?

hemasekhar-p:
@damianmomotgoogle, I have Refactored the code as per the review comments. could you please review?

dev/src/main/java/com/google/adk/web/config/OpenTelemetryConfig.java
hemasekhar-p: Refactored the code as per the review comment.

damianmomotgoogle (approved):

PR #1185, opened 2026-05-06
fix: Resolve IllegalArgumentException for text MIME types in LangChain4j adapter

### LINKED ISSUE #1184: `IllegalArgumentException: "Unknown or unhandled mime type"` incorrectly thrown for valid text/JSON/XML inputs in `LangChain4j` adapter
### Description

There is a logic error in `com.google.adk.models.langchain4j.LangChain4j` within the `toUserOrToolResultMessage(Content content)` method. When parsing `text/*`, `application/json`, or `+xml` mime types, the code correctly creates a `TextContent` object but adds it directly to the `lc4jContents` list while leaving the local variable `lc4jContent` as `null`. 

Immediately after this `if/else if` block, the code checks if `lc4jContent == null` and throws an `IllegalArgumentException`, making it impossible to process these standard text and JSON/XML content types.

### Environment
* **ADK Version:** 1.2.0
* **LangChain4J Version:** 1.14.0
* **Class:** `com.google.adk.models.langchain4j.LangChain4j`

### Steps to Reproduce
Pass any `Content` object containing an `inlineData` Part with a text-based mime type (e.g., `text/plain`, `application/json`) through the LangChain4J model adapter.

### Root Cause Analysis
In `toUserOrToolResultMessage`, the `if/else` block handles different mime types:
```java
dev.langchain4j.data.message.Content lc4jContent = null;

// ... [audio/video/image/pdf logic successfully assigns lc4jContent] ...

} else if (mimeType.startsWith("text/") || "application/json".equals(mimeType) || mimeType.endsWith("+json") || mimeType.endsWith("+xml")) {
    // ❌ BUG: Adds directly to the list, leaving lc4jContent as null
    lc4jContents.add(TextContent.from(new String(bytes, StandardCharsets.UTF_8)));
}

// ❌ Throws exception because lc4jContent is still null for text/json/xml
if (lc4jContent == null) {
    throw new IllegalArgumentException("Unknown or unhandled mime type: " + mimeType);
}

// (For other types, it gets added here)
lc4jContents.add(lc4jContent);
```

### Expected Behavior
Text-based content should be parsed into a `TextContent` object and successfully added to the message list without throwing an `IllegalArgumentException`.

### Suggested Fix
Update the `else if` branch for text/JSON types to assign the value to `lc4jContent` instead of adding it directly to the list, maintaining consistency with the other mime type branches:

```java
} else if (mimeType.startsWith("text/") || "application/json".equals(mimeType) || mimeType.endsWith("+json") || mimeType.endsWith("+xml")) {
    // ✅ FIX: Assign to the local variable instead
    lc4jContent = TextContent.from(new String(bytes, StandardCharsets.UTF_8));
}

if (lc4jContent == null) {
    throw new IllegalArgumentException("Unknown or unhandled mime type: " + mimeType);
}

lc4jContents.add(lc4jContent);
```

hemasekhar-p:

- Closes: #1184

**Problem:**
When `LangChain4j.toUserOrToolResultMessage()` processes `inlineData` parts with text-based MIME types, it instantiates a `TextContent` object but adds it directly to the master list rather than assigning it to the local `lc4jContent` reference. this is causing the false validation for this condition `if (lc4jContent == null)` and throwing an unexpected `IllegalArgumentException`. Additionally, the existing string matching was too strict `.equals()`, causing valid payloads to fail if a charset parameter was appended like this `application/json; charset=utf-8`

**Solution:**
Modified the text MIME type matching logic to use `.startsWith()` and `.contains()` to gracefully tolerate appended parameters like boundaries or charsets.

Implemented a private `extractCharset()` helper method to dynamically read the provided character set and safely falling back to `UTF-8` if no charset provided or malformed.

Updated the control flow to assign the generated `TextContent` directly to the `lc4jContent` variable,

### Testing Plan

_Please describe the tests that you ran to verify your changes. This is required
for all PRs that are not small documentation or typo fixes._

**Unit Tests:**

Added the following methods to `LangChain4jTest.java` which all execute and pass successfully:

`testGenerateContentWithTextPlainInlineData` - Verifies standard text parsing.

`testGenerateContentWithApplicationJsonInlineData` - Verifies JSON payload parsing.

`testGenerateContentWithUnsupportedMimeType` - Verifies the unsupported types still correctly throws IllegalArgumentException for truly unhandled types.

`testGenerateContentWithExplicitCharset` - Verifies the dynamic extraction of custom charsets.

`testGenerateContentWithMalformedCharsetFallback` - Verifies safe fallback to UTF-8 when provided with unsupported charset parameters.

### DIFF
--- a/contrib/langchain4j/src/main/java/com/google/adk/models/langchain4j/LangChain4j.java
+++ b/contrib/langchain4j/src/main/java/com/google/adk/models/langchain4j/LangChain4j.java
@@ -70,6 +70,7 @@ import dev.langchain4j.model.chat.response.StreamingChatResponseHandler;
 import dev.langchain4j.model.output.TokenUsage;
 import io.reactivex.rxjava3.core.BackpressureStrategy;
 import io.reactivex.rxjava3.core.Flowable;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.HashMap;
@@ -77,10 +78,13 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import org.jspecify.annotations.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @AutoValue
 public abstract class LangChain4j extends BaseLlm {
 
+  private static final Logger logger = LoggerFactory.getLogger(LangChain4j.class);
   private static final TypeReference<Map<String, Object>> MAP_TYPE_REFERENCE =
       new TypeReference<>() {};
 
@@ -388,13 +392,12 @@ public abstract class LangChain4j extends BaseLlm {
                       .mimeType(mimeType)
                       .build());
         } else if (mimeType.startsWith("text/")
-            || "application/json".equals(mimeType)
-            || mimeType.endsWith("+json")
-            || mimeType.endsWith("+xml")) {
+            || mimeType.startsWith("application/json")
+            || mimeType.contains("+json")
+            || mimeType.contains("+xml")) {
           // TODO are there missing text based mime types?
           // TODO should we assume UTF_8?
-          lc4jContents.add(
-              TextContent.from(new String(bytes, java.nio.charset.StandardCharsets.UTF_8)));
+          lc4jContent = TextContent.from(new String(bytes, extractCharset(mimeType)));
         }
 
         if (lc4jContent != null) {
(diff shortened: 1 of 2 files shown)


hemasekhar-p:
@krwc Please take a look into it.

contrib/langchain4j/src/main/java/com/google/adk/models/langchain4j/LangChain4j.java
```diff
@@ -420,6 +420,20 @@ private List<ChatMessage> toUserOrToolResultMessage(Content content) {
...
+    String charSetString = "charset=";
+    if (mimeType == null || !mimeType.toLowerCase().contains(charSetString)) {
+      return java.nio.charset.StandardCharsets.UTF_8;
+    }
+    try {
+      String[] parts = mimeType.toLowerCase().split(charSetString);
+      String charsetName = parts[1].split(";")[0].trim().replace("\"", "").replace("'", "");
+      return java.nio.charset.Charset.forName(charsetName);
+    } catch (IllegalArgumentException e) {
+      return java.nio.charset.StandardCharsets.UTF_8;
```
damianmomotgoogle, line 438: Could we add warn log?

Otherwise unexpected cases might go completely unnoticed

damianmomotgoogle (approved):

contrib/langchain4j/src/main/java/com/google/adk/models/langchain4j/LangChain4j.java
hemasekhar-p: Added the worn log. could you please review now?

damianmomotgoogle (approved):

PR #1222, opened 2026-05-26
fix(dev): restrict default CORS origins and WebSocket allowed origins to localhost

g0w6y:
## Problem

The dev server (`AdkWebServer` / `adk web`) ships two independent wildcard origin policies:

1. **HTTP CORS** – `AdkWebCorsProperties` falls back to `["*"]` when no origins are configured, so every response carries `Access-Control-Allow-Origin: *`. Any page on the internet can therefore read agent responses cross-origin.

2. **WebSocket (`/run_live`)** – `WebSocketConfig` calls `.setAllowedOrigins("*")`, making the endpoint accept upgrade requests from any `Origin` header. This allows Cross-Site WebSocket Hijacking (CSWSH): a malicious page can open a live session against a dev server that is reachable from the victim's browser.

Combined, these let a remote origin read HTTP responses and drive the agent over WebSocket without any user interaction beyond visiting a page.

## Fix

**`AdkWebCorsProperties.java`** – change the default `origins` fallback from `["*"]` to `["http://localhost:8080", "http://127.0.0.1:8080"]`. The dev UI (served on the same host) continues to work; all other origins are blocked by the browser.

**`WebSocketConfig.java`** – inject `AdkWebCorsProperties` and derive `setAllowedOrigins` from the same property instead of a separate hardcoded wildcard, so both policies stay in sync. Users who need a broader origin allowlist can set `adk.web.cors.origins` explicitly in their application properties.

## Impact

- No change to users running the standard dev workflow (UI on `localhost:8080`).
- Users who access the dev server from a custom host/port need to add that origin via `adk.web.cors.origins`.
- Eliminates wildcard CORS and CSWSH exposure on `/run_live` for all default-configuration deployments.

### DIFF
--- a/dev/src/main/java/com/google/adk/web/config/AdkWebCorsProperties.java
+++ b/dev/src/main/java/com/google/adk/web/config/AdkWebCorsProperties.java
@@ -37,7 +37,7 @@ public record AdkWebCorsProperties(
     origins =
         origins != null && !origins.isEmpty()
             ? origins
-            : List.of("http://localhost:8080", "http://127.0.0.1:8080");
+            : List.of("http://localhost:[*]", "http://127.0.0.1:[*]");
     methods =
         methods != null && !methods.isEmpty()
             ? methods
--- a/dev/src/main/java/com/google/adk/web/config/AdkWebCorsProperties.java
+++ b/dev/src/main/java/com/google/adk/web/config/AdkWebCorsProperties.java
@@ -34,10 +34,7 @@ public record AdkWebCorsProperties(
 
   public AdkWebCorsProperties {
     mapping = mapping != null ? mapping : "/**";
-    origins =
-        origins != null && !origins.isEmpty()
-            ? origins
-            : List.of("http://localhost:[*]", "http://127.0.0.1:[*]");
+    origins = origins != null && !origins.isEmpty() ? origins : List.of("*");
     methods =
         methods != null && !methods.isEmpty()
             ? methods
--- a/dev/src/main/java/com/google/adk/web/config/AdkWebCorsProperties.java
+++ b/dev/src/main/java/com/google/adk/web/config/AdkWebCorsProperties.java
@@ -34,7 +34,10 @@ public record AdkWebCorsProperties(
 
   public AdkWebCorsProperties {
     mapping = mapping != null ? mapping : "/**";
-    origins = origins != null && !origins.isEmpty() ? origins : List.of("*");
+    origins =
+        origins != null && !origins.isEmpty()
+            ? origins
+            : List.of("http://localhost:8080", "http://127.0.0.1:8080");
     methods =
         methods != null && !methods.isEmpty()
             ? methods
--- a/dev/src/main/java/com/google/adk/web/config/AdkWebCorsConfig.java
+++ b/dev/src/main/java/com/google/adk/web/config/AdkWebCorsConfig.java
@@ -46,7 +46,7 @@ public class AdkWebCorsConfig {
   public CorsConfigurationSource corsConfigurationSource(AdkWebCorsProperties corsProperties) {
     CorsConfiguration configuration = new CorsConfiguration();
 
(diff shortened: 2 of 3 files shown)


hemasekhar-p:
Hi @g0w6y, Thank you for your contribution! We appreciate you taking the time to submit this pull request. Currently this PR is under review by our team, we will keep you posted if any additional information is required. thank you.

hemasekhar-p:

dev/src/main/java/com/google/adk/web/config/AdkWebCorsProperties.java
```diff
@@ -34,7 +34,10 @@ public record AdkWebCorsProperties(
 
   public AdkWebCorsProperties {
     mapping = mapping != null ? mapping : "/**";
-    origins = origins != null && !origins.isEmpty() ? origins : List.of("*");
+    origins =
+        origins != null && !origins.isEmpty()
+            ? origins
+            : List.of("http://localhost:8080", "http://127.0.0.1:8080");
```
krwc: This is locking us to 8080 port, which may not be the port that the server is running on. 

I see two options:
1) We use `:[*]`.
2) We pass the actual port, but then we'll stumble across a problem when user selects an ephemeral port 0 (in a very unlikely scenario).

krwc (changes requested):

dev/src/main/java/com/google/adk/web/config/AdkWebCorsProperties.java
g0w6y: Went with option 1. The default origins are now the localhost port wildcard (`http://localhost:[*]`, `http://127.0.0.1:[*]`), so any local port works while remote origins stay blocked. Both sinks now use `setAllowedOriginPatterns`, since the `[*]` pattern is only honored there.

krwc:
Thanks for iterating on this @g0w6y, and sorry for the extra round-trip.

We've discussed this with the team, and realized that if we changed the default origins now to a fixed list of hosts, we would potentially break existing customers' workflow. Admittedly, `*` is not the best choice as a default, and should be changed in the future major release (we track this internally now), however for now, we'd like to retain backwards compatibility.

As such, I'd propose to do the following:
1. Let's keep the default as `*` (no change wrt to HEAD)
1. Let's also use the origins for websockets (new behaviour).
1. Let's add a warning log that using `*` is not secure by default and that it should be changed via origins field by the user.

While it doesn't address the originally stated goal, it leaves a way for the user to configure it correctly, so it's still an improvement compared to the current state.

g0w6y:
Done, updated per the team's decision:

1. Default origins kept as `*` (no change vs HEAD).
2. `/run_live` now derives its allowed origins from `adk.web.cors.origins` instead of a hardcoded `*`, so both HTTP CORS and the WebSocket are controlled by the single property.
3. Added a startup WARN when origins is `*`, flagging it as insecure (dev only) and pointing users to set `adk.web.cors.origins`.

Net diff is 2 files and default behaviour is unchanged. Verified against Boot 4.0.2 / Spring 7.0.3: with the default the WARN fires and `*` still allows all origins; with an explicit list (e.g. `http://localhost:4200`) there is no warning and remote origins are blocked.

g0w6y:
Hi @krwc, could you please review the latest changes when you have a moment?

Also, would the team consider publishing a GitHub Security Advisory for this, so the wildcard CORS and WebSocket origin exposure is tracked publicly? Happy to help with the writeup if useful. Thank you.

krwc (approved):

krwc:
Hi @g0w6y! Thanks for updating the PR. Should get merged soon :-)

Answering your question: the intent behind ADK Web Server is to be for development purposes only ([link](https://adk.dev/runtime/web-interface/#use-the-web-interface), direct quote: _ADK Web is **not meant for use in production deployments**. You should use ADK Web for development and debugging purposes only_).  

As such we're not convinced it deserves Github Security Advisory. LMK what you think.

PR #1228, opened 2026-05-26
fix(skills): prevent path traversal in LocalSkillSource

adilburaksen:
## Summary

`LocalSkillSource` resolves caller-supplied `skillName` and `resourcePath`
strings directly onto the file system without validating that the resulting
path remains within `skillsBasePath`. A crafted value such as
`../../../etc/passwd` passes the `Files.exists()` check (the OS `stat` call
follows `..` segments) and is returned as a valid resource path, enabling
arbitrary file reads from the host.

Affected methods: `findResourcePath`, `listResources`, `findSkillMdPath`.

## Fix

Added a private `validatePathWithinBase(Path base, String component)` helper
that rejects:

* absolute path components (e.g. `/etc/passwd`)
* components whose normalized, absolute resolution escapes `base`
  (e.g. `../../secret`)

The check mirrors the boundary enforcement already present in the Go
implementation (`filesystem_source.go`).

## Tests

Five new test cases in `LocalSkillSourceTest`:

| Test | Input | Expected |
|------|-------|----------|
| `testLoadResource_pathTraversalInSkillName` | skillName `../other-dir` | `SkillSourceException` "Path traversal detected" |
| `testLoadResource_pathTraversalInResourcePath` | resourcePath `../../../etc/passwd` | `SkillSourceException` "Path traversal detected" |
| `testLoadResource_absoluteResourcePath` | resourcePath `/etc/passwd` | `SkillSourceException` "Absolute paths are not allowed" |
| `testListResources_pathTraversalInSkillName` | skillName `../../etc` | `SkillSourceException` "Path traversal detected" |
| `testListResources_pathTraversalInResourceDirectory` | resourceDirectory `../other-skill` | `SkillSourceException` "Path traversal detected" |


hemasekhar-p:
Hi @adilburaksen, thank you for your contribution! We appreciate you taking the time to submit this pull request. I noticed some formatting inconsistencies in the test cases. could you please address these issues and run the full Maven build without skipping the test cases to ensure everything is compliant?

hemasekhar-p:
@adilburaksen, Thank you for your quick response. As per contribution policy, please ensure your PR consists of a single commit. Could you please change your commits accordingly?

adilburaksen:
Thanks for the review @hemasekhar-p. I've applied google-java-format to the test file so it matches the project style now, and I squashed everything down into a single commit as requested.

I also ran the full build with tests enabled (no skips) and it passes, including LocalSkillSourceTest (16 tests, all green). Let me know if anything else is needed.

hemasekhar-p:
Thank you for your update @adilburaksen, Currently this PR is under review by our team, we will keep you posted if any additional information is required. thank you.

hemasekhar-p:

adilburaksen:
Rebased onto current `main` — this is now a clean 2-file diff (the earlier large diff was a stale-branch artifact from an old base). `validatePathWithinBase` rejects absolute paths and any `..` escape, applied in `findResourcePath`, `listResources`, and `findSkillMdPath`, with tests covering skillName and resourceDirectory traversal. PTAL @sherryfox @hemasekhar-p — happy to adjust.

adilburaksen:
Quick follow-up: the Google OSS VRP report tracking this is now on hold pending an official **merge**, so a review here would really help close it out.

The PR is ready from my side — single commit, CI green (CLA + check-changes), no conflicts with `main`, and `validatePathWithinBase` rejects absolute paths and `..` escapes in `findResourcePath`, `listResources`, and `findSkillMdPath`, with tests covering both `skillName` and `resourceDirectory` traversal.

@sherryfox @hemasekhar-p — PTAL whenever you get a chance, happy to make any changes.

adilburaksen:
Friendly follow-up on this one. CI is green (CLA + check-changes passing), it's a single clean commit scoped to `LocalSkillSource` path validation with tests.

Since my last note, the Google OSS VRP panel confirmed (2026-06-26) that they'll proceed with the reward evaluation **immediately once this PR is merged** — the report is otherwise ready and only gated on the upstream merge. @sherryfox @hemasekhar-p would you be able to take a look / import when you have a moment? Happy to rebase or adjust anything needed. Thanks!

core/src/main/java/com/google/adk/skills/LocalSkillSource.java
```diff
@@ -45,6 +45,12 @@ public LocalSkillSource(Path skillsBasePath) {
 
   @Override
   public Single<ImmutableList<String>> listResources(String skillName, String resourceDirectory) {
+    try {
+      validatePathWithinBase(skillsBasePath, skillName);
+      validatePathWithinBase(skillsBasePath.resolve(skillName), resourceDirectory);
+    } catch (SkillSourceException e) {
+      return Single.error(e);
+    }
     Path skillDir = skillsBasePath.resolve(skillName);
```
wikaaaaa: Can we re-use the normalized path computed by validatePathWithBase?  (here and in all other places its called)

We compute the normalized path, throw it away and resolve again, it would be better if we used the same path we validated.

core/src/test/java/com/google/adk/skills/LocalSkillSourceTest.java
```diff
@@ -374,4 +374,31 @@ public void testLoadInstructions_emptyFile() throws IOException {
...
+  @Test
+  public void testListResources_pathTraversalRejected() throws IOException {
+    Path skillsBase = tempFolder.getRoot().toPath().resolve("skills");
+    Files.createDirectory(skillsBase);
+
+    SkillSource source = new LocalSkillSource(skillsBase);
+
+    var single = source.listResources("../../etc", "");
+    RuntimeException exception = assertThrows(RuntimeException.class, single::blockingGet);
+    assertThat(exception).hasCauseThat().isInstanceOf(SkillSourceException.class);
```
wikaaaaa, line 405: This test passes with or without the fix, so it doesn't actually guard the regression. listResources("../../etc", "") resolves to a non-existent directory, so the pre-existing isDirectory check already returns SkillSourceException("Skill not found") — and the test only asserts isInstanceOf(SkillSourceException.class). 

Let's assert on the traversal error message (here and in other test cases as well).

core/src/main/java/com/google/adk/skills/LocalSkillSource.java
```diff
@@ -122,6 +139,22 @@ protected ReadableByteChannel openChannel(Path path) throws IOException {
     return Files.newByteChannel(path);
   }
 
+  private static void validatePathWithinBase(Path base, String component)
+      throws SkillSourceException {
+    if (Path.of(component).isAbsolute()) {
```
wikaaaaa: wdyt about changing it to 
`if (base.getFileSystem().getPath(component).isAbsolute()) { ... }`?

Path.of(component) always parses against FileSystems.getDefault(), but base.resolve(component) two lines down parses against base's filesystem. LocalSkillSource accepts an arbitrary Path skillsBasePath, which may come from a non-default provider.

core/src/main/java/com/google/adk/skills/LocalSkillSource.java
```diff
@@ -122,6 +139,22 @@ protected ReadableByteChannel openChannel(Path path) throws IOException {
...
+      throw new SkillSourceException(
+          "Absolute paths are not allowed: " + component, SKILL_NOT_FOUND);
```
wikaaaaa: nit: the helper is used for validating both skill and resource paths, but throws SKILL_NOT_FOUND for both. RESOURCE_NOT_FOUND would be more consistent for the resource-path cases — though since it's a shared helper, this'd mean passing the error code in (or splitting it) so it can tell the two apart.

core/src/main/java/com/google/adk/skills/LocalSkillSource.java
```diff
@@ -122,6 +139,22 @@ protected ReadableByteChannel openChannel(Path path) throws IOException {
...
+    }
+    Path normalizedBase = base.normalize().toAbsolutePath();
+    Path resolved = base.resolve(component).normalize().toAbsolutePath();
+    if (!resolved.startsWith(normalizedBase)) {
+      throw new SkillSourceException(
+          "Path traversal detected; component must remain within its parent directory: "
```
wikaaaaa: nit: "must remain within its parent directory" is imprecise — base is the containment root, which is the skills base dir for the skillName checks but the individual skill dir for the resource-path checks. Use a generic noun ("must not escape its base directory"), or if the helper becomes context-aware for the error code, say "skills base directory" vs "skill directory" accordingly. (And keep base out of the message to avoid leaking the absolute path.)

core/src/main/java/com/google/adk/skills/LocalSkillSource.java
```diff
@@ -106,6 +118,11 @@ protected Single<Path> findResourcePath(String skillName, String resourcePath) {
 
   @Override
   protected Single<Path> findSkillMdPath(String skillName) {
```
wikaaaaa, line 120: findSkillMdPath now has a skill_name guard but no test, please add a test case for that.

wikaaaaa (changes requested):
Thank you for the fix!

The approach looks great, I left some comments.

core/src/main/java/com/google/adk/skills/LocalSkillSource.java
wikaaaaa: Actually I can see you listed 5 new test cases in the description, but I can only see 2 in the changes. Can you please make sure all your changes are uploaded?

adilburaksen:
Thanks for the detailed review! Addressed all points in the latest push:

- **Reuse the validated path:** `validatePathWithinBase` now returns the normalized/validated `Path`, and `listResources`, `findResourcePath` and `findSkillMdPath` use it directly instead of re-resolving `skillsBasePath.resolve(skillName)` a second time.
- **Filesystem-aware absolute check:** switched `Path.of(component)` → `base.getFileSystem().getPath(component)` so the parse matches the filesystem `base.resolve()` uses (handles a non-default-provider `skillsBasePath`).
- **Error code:** the helper now takes the error code, so resource-path checks report `RESOURCE_NOT_FOUND` and skill-name checks report `SKILL_NOT_FOUND`.
- **Message:** reworded to "component must not escape its base directory" (kept `base` out of the message).
- **Tests:** the two existing traversal tests now assert on the traversal error message (previously they passed with or without the fix), and I added the missing cases — `findSkillMdPath` (via `loadFrontmatter`), resource-path traversal, and absolute-path rejection. All five assert on the specific error message now.

`mvn -pl core test -Dtest=LocalSkillSourceTest` passes locally.

core/src/main/java/com/google/adk/skills/LocalSkillSource.java
```diff
@@ -122,6 +139,24 @@ protected ReadableByteChannel openChannel(Path path) throws IOException {
     return Files.newByteChannel(path);
   }
 
+  private static Path validatePathWithinBase(Path base, String component, String errorCode)
+      throws SkillSourceException {
+    // Parse the component against the base's own filesystem: LocalSkillSource accepts an arbitrary
+    // skillsBasePath, which may come from a non-default provider, so Path.of (which always uses
+    // FileSystems.getDefault()) could disagree with base.resolve below.
```
wikaaaaa: I think we can skip this comment. Or not mention Path.of() since its not being used, mentioning it feels unnecessary.

core/src/test/java/com/google/adk/skills/LocalSkillSourceTest.java
```diff
@@ -374,4 +374,83 @@ public void testLoadInstructions_emptyFile() throws IOException {
...
+    Files.writeString(tempFolder.getRoot().toPath().resolve("secret.txt"), "top-secret");
+
+    SkillSource source = new LocalSkillSource(skillsBase);
+
+    // A skill name that escapes the skills base via "..".
+    var single = source.loadResource("..", "secret.txt");
+    RuntimeException exception = assertThrows(RuntimeException.class, single::blockingGet);
+    assertThat(exception).hasCauseThat().isInstanceOf(SkillSourceException.class);
+    assertThat(exception).hasCauseThat().hasMessageThat().contains("Path traversal detected");
+  }
```
wikaaaaa, line 394: lets assert the error codes as well (in all test cases).

wikaaaaa (changes requested):

wikaaaaa:
One of the checks is failing for this PR:

"This pull request has 2 commits.
Please squash them into a single commit before merging."

adilburaksen:
Done — squashed into a single commit (content unchanged). Ready for another look.

wikaaaaa:
Thanks for the update! The GitHub checks pass now. Just the two remaining review comments to address (error-code assertions + the Path.of comment tweak) and we can merge.

adilburaksen:
Superseded by #1331 — this PR's branch was accidentally force-pushed and GitHub wouldn't allow reopening. The identical fix, with all review feedback applied, continues in #1331. Apologies for the noise.

adilburaksen:
@wikaaaaa heads-up — since this branch can't be reopened, the same fix with both of your review comments applied continues in #1331. Would appreciate you picking up the review there. Thanks, and sorry for the churn.

PR #1255, opened 2026-06-09
feat: Add injectable time and UUID providers

DABH:

**Problem:**
ADK generates timestamps and IDs by calling `Instant.now()` and `UUID.randomUUID()` directly. There are use cases where it's useful to have custom clock and UUID providers, such as when you want to make runs reproducible/deterministic.

`adk-python` already solves this with `platform.time` / `platform.uuid` - see https://github.com/google/adk-python/pull/4200.  Closely inspired by that, we propose adding similar functionality for `adk-java`.  Note that a similar PR is open right now for `adk-go`: https://github.com/google/adk-go/pull/964.

**Solution:**
Introduce a leaf package `com.google.adk.platform` with two functional interfaces:

- `TimeProvider` — `Instant now()`, default `SYSTEM` backed by `Instant::now`.
- `UuidProvider` — `String newUuid()`, default `SYSTEM` backed by `UUID.randomUUID()`.

Both default to today's wall-clock / random behavior, so this change is fully backwards-compatible.

The providers are **threaded as data through `InvocationContext`** rather than stored in an ambient `ThreadLocal`. ADK's core flow hops threads via `RxJava` (e.g. `BaseLlmFlow` observes on `Schedulers.io()` / the agent executor), and events are built downstream on those worker threads. We considered using a `ThreadLocal` but that would silently fall back to the system providers after a thread hop, failing open with no error. So the proposed solution here, carrying the providers on the `InvocationContext` (which is already captured in the RxJava lambdas and passed down the chain) makes them visible on whatever thread builds an event, so this is thread-safe. This mirrors the proposed changes to `adk-go`, which diverged from `adk-python`'s contextvar mechanism for the same concurrency reason.

Users configure the providers **once** on the `Runner` (`Runner.builder().timeProvider(...).uuidProvider(...)`); no knowledge of ADK's internal threads is required.

The other files edited in this PR are just call sites updated to derive from the in-scope providers:
- `InvocationContext` — invocation ID, plus `now()` / `newUuid()` accessors.
- `BaseLlmFlow`, `Functions`, `OutputSchema` — event IDs + timestamps, function-call IDs.
- `Runner` — invocation ID + event builds.
- `InMemorySessionService` — auto session ID + `lastUpdateTime` (constructor injection; default constructor keeps `SYSTEM` behavior).

`Event.generateEventId()` and the `Event.Builder.build()` timestamp default now delegate to `UuidProvider.SYSTEM` / `TimeProvider.SYSTEM`, so the platform interfaces are the single source for generated IDs and timestamps while the `Event` API stays unchanged.

### Testing Plan

**Unit Tests:**

New/updated tests:
- `platform/TimeProviderTest`, `platform/UuidProviderTest` — `SYSTEM` defaults + custom-provider behavior.
- `InvocationContextTest` — providers default to `SYSTEM`, are carried by the builder/`toBuilder()`, and drive `now()` / `newUuid()`.
- `FunctionsTest` — function-call ID derives from the injected provider.
- `InMemorySessionServiceTest` — injected providers produce a deterministic session ID and `lastUpdateTime`.
- `RunnerTest.runAsync_withDeterministicProviders_producesIdenticalEvents` — the headline guarantee: running the same input twice through the full flow yields byte-identical event IDs, timestamps, and invocation IDs.

Local results (`./mvnw -pl core test`, openjdk@17), all green:
```
TimeProviderTest / UuidProviderTest  Tests run: 2,   Failures: 0, Errors: 0, Skipped: 0
RunnerTest                            Tests run: 139, Failures: 0, Errors: 0, Skipped: 0
InvocationContextTest / FunctionsTest / InMemorySessionServiceTest  Failures: 0, Errors: 0
```
`./mvnw -pl core fmt:check` — 0 non-complying files.

This is an in-process library seam with no runtime/config surface to exercise manually. The guarantee is verified by the unit test above, which asserts two independent runs of the same invocation produce identical IDs and timestamps.

### Checklist

- [x] I have read the CONTRIBUTING.md document.

### Additional context

Motivated by maintaining feature parity with `adk-python` and the proposed changes to `adk-go`.

### DIFF
--- a/core/src/main/java/com/google/adk/agents/InvocationContext.java
+++ b/core/src/main/java/com/google/adk/agents/InvocationContext.java
@@ -22,6 +22,8 @@ import com.google.adk.apps.ResumabilityConfig;
 import com.google.adk.artifacts.BaseArtifactService;
 import com.google.adk.memory.BaseMemoryService;
 import com.google.adk.models.LlmCallsLimitExceededException;
+import com.google.adk.platform.TimeProvider;
+import com.google.adk.platform.UuidProvider;
 import com.google.adk.plugins.Plugin;
 import com.google.adk.plugins.PluginManager;
 import com.google.adk.sessions.BaseSessionService;
@@ -29,10 +31,10 @@ import com.google.adk.sessions.Session;
 import com.google.adk.summarizer.EventsCompactionConfig;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import com.google.genai.types.Content;
+import java.time.Instant;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.jspecify.annotations.Nullable;
@@ -56,6 +58,8 @@ public class InvocationContext {
   private final @Nullable ResumabilityConfig resumabilityConfig;
   private final InvocationCostManager invocationCostManager;
   private final Map<String, Object> callbackContextData;
+  private final TimeProvider timeProvider;
+  private final UuidProvider uuidProvider;
 
   @Nullable private String branch;
   private BaseAgent agent;
@@ -83,6 +87,8 @@ public class InvocationContext {
     // invocation invocation so that Plugins can access the same data it during the invocation
     // across all types of callbacks.
     this.callbackContextData = builder.callbackContextData;
+    this.timeProvider = builder.timeProvider;
+    this.uuidProvider = builder.uuidProvider;
   }
 
   /** Returns a new {@link Builder} for creating {@link InvocationContext} instances. */
@@ -197,9 +203,34 @@ public class InvocationContext {
     return session.userId();
   }
 
+  /** Returns the {@link TimeProvider} for this invocation. */
+  public TimeProvider timeProvider() {
+    return timeProvider;
+  }
+
+  /** Returns the {@link UuidProvider} for this invocation. */
+  public UuidProvider uuidProvider() {
+    return uuidProvider;
+  }
+
+  /** Returns the current time from this invocation's {@link TimeProvider}. */
+  public Instant now() {
+    return timeProvider.now();
+  }
+
+  /** Returns a new unique identifier from this invocation's {@link UuidProvider}. */
+  public String newUuid() {
+    return uuidProvider.newUuid();
+  }
+
   /** Generates a new unique ID for an invocation context. */
   public static String newInvocationContextId() {
-    return "e-" + UUID.randomUUID();
+    return newInvocationContextId(UuidProvider.SYSTEM);
+  }
+
+  /** Generates a new unique ID for an invocation context using the given {@link UuidProvider}. */
+  public static String newInvocationContextId(UuidProvider uuidProvider) {
+    return "e-" + uuidProvider.newUuid();
   }
 
   /**
@@ -289,6 +320,8 @@ public class InvocationContext {
       // invocation invocation so that Plugins can access the same data it during the invocation
       // across all types of callbacks.
       this.callbackContextData = context.callbackContextData;
+      this.timeProvider = context.timeProvider;
+      this.uuidProvider = context.uuidProvider;
     }
 
     private BaseSessionService sessionService;
@@ -309,6 +342,8 @@ public class InvocationContext {
     private @Nullable ResumabilityConfig resumabilityConfig;
     private InvocationCostManager invocationCostManager = new InvocationCostManager();
     private Map<String, Object> callbackContextData = new ConcurrentHashMap<>();
+    private TimeProvider timeProvider = TimeProvider.SYSTEM;
+    private UuidProvider uuidProvider = UuidProvider.SYSTEM;
 
     /**
      * Sets the session service for managing session state.
@@ -503,6 +538,30 @@ public class InvocationContext {
     }
 
     /**
+     * Sets the time provider for the invocation. Defaults to {@link TimeProvider#SYSTEM}.
+     *
+     * @param timeProvider the provider for the current time.
+     * @return this builder instance for chaining.
+     */
+    @CanIgnoreReturnValue
+    public Builder timeProvider(TimeProvider timeProvider) {
+      this.timeProvider = timeProvider;
+      return this;
+    }
+
+    /**
+     * Sets the UUID provider for the invocation. Defaults to {@link UuidProvider#SYSTEM}.
+     *
+     * @param uuidProvider the provider for new unique identifiers.
+     * @return this builder instance for chaining.
+     */
+    @CanIgnoreReturnValue
+    public Builder uuidProvider(UuidProvider uuidProvider) {
+      this.uuidProvider = uuidProvider;
+      return this;
+    }
+
+    /**
      * Builds the {@link InvocationContext} instance.
      *
      * @throws IllegalStateException if any required parameters are missing.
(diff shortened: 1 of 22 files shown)


hemasekhar-p:
Hi @DABH, thank you for your contribution! We appreciate you taking the time to submit this pull request. As per contribution policy, please ensure your PR consists of a single commit and i have noticed inconsistency code formatting. Could you please change your commits accordingly and address the code formatting issue?

core/src/main/java/com/google/adk/agents/InvocationContext.java
```diff
@@ -21,17 +21,19 @@
 import com.google.adk.artifacts.BaseArtifactService;
```
xumaple, line 22: BaseAgent.java:421 and :438 still build events with Event.generateEventId()

(according to claude) - should they be swapped out as well?

core/src/main/java/com/google/adk/flows/llmflows/BaseLlmFlow.java
```diff
@@ -438,7 +438,7 @@ private Flowable<Event> runOneStep(Context spanContext, InvocationContext contex
 
                         final Event mutableEventTemplate =
                             Event.builder()
-                                .id(Event.generateEventId())
```
xumaple, line 454: What is the rationale to swap out all of these calls verses just replacing the one generateEventId() function impl with the custom impl?

xumaple: nvm, see it in the description. But not sure I agree with the premise. If the default implementation remains the same, doing the entire swap should not be a breaking change at all. (And tests should be able to ensure such)

core/src/test/java/com/google/adk/runner/RunnerTest.java
```diff
@@ -1531,6 +1533,68 @@ public void runAsync_ensureEventsAreAppendedInOrder() throws Exception {
...
+          public String newUuid() {
+            return String.format("uuid-%04d", counter.getAndIncrement());
+          }
+        };
+    LlmAgent deterministicAgent =
+        createTestAgentBuilder(createTestLlm(createLlmResponse(createContent("from llm")))).build();
+    Runner deterministicRunner =
+        Runner.builder()
+            .app(App.builder().name("test").rootAgent(deterministicAgent).build())
+            .sessionService(new InMemorySessionService(fixedClock, sequentialUuids))
```
xumaple: The fact that we have to wire this in two places (lines 1583:1585) suggests that something is not quite right here to me. Perhaps consider having Runner.build() inject its providers into the session service when it owns/constructs it?

core/src/main/java/com/google/adk/agents/InvocationContext.java
xumaple: Claude also flagged: core/.../summarizer/LlmEventSummarizer.java:105 builds the compaction event with .invocationId(Event.generateEventId()) and no .id(...)

xumaple, line 22: TraceManager, BigQueryAgentAnalyticsPlugin, PluginState, Parser still call Instant.now()/UUID.randomUUID(). These are telemetry related - so a bigger change. We should only change them if we expect to build a plugin which actually exercises this path? What do you think?

hemasekhar-p:
@DABH, thank you for your quick response. when running a Maven build, some files are modifying due to incorrect formatting. so, use the Maven build to ensure your changes are properly aligned and please make sure your PR contains a single commit.

core/src/main/java/com/google/adk/agents/InvocationContext.java
DABH: Good catch — swept these. BaseAgent's two agent-callback event builds, the three `CodeExecution` event builds, and the `LlmEventSummarizer` compaction event now mint their id and timestamp from the in-scope providers. The summarizer receives the providers via constructor injection from the `Runner`, and its compaction event now also sets an explicit `.id(...)` (previously it set only `invocationId`).

core/src/main/java/com/google/adk/flows/llmflows/BaseLlmFlow.java
DABH: Agreed, and done. I completed the full call-site sweep, and to make the default behavior provably unchanged I also routed `Event.generateEventId()` and the `Event.Builder` timestamp default through `UuidProvider.SYSTEM` / `TimeProvider.SYSTEM`, so the platform interfaces are now the single source for generated ids/timestamps.

I kept the per-call-site threading (rather than only swapping the static impl) because the override has to ride on the `InvocationContext`: a process-global static or `ThreadLocal` can't isolate providers to a single run — two `Runner`s in one JVM would interfere — and a `ThreadLocal` silently reverts to the wall clock after the RxJava `observeOn` thread hop. The `RunnerTest` case asserts byte-identical output across two identical runs to lock the non-breaking guarantee in.

core/src/test/java/com/google/adk/runner/RunnerTest.java
DABH: Fixed — `Runner.build()` now constructs its default `InMemorySessionService` with the runner's providers, so callers wire providers once on the `Runner`. The test no longer creates a separate session service. A user-supplied `BaseSessionService` stays user-configured, as before.

core/src/main/java/com/google/adk/agents/InvocationContext.java
DABH: Agreed — leaving those as-is for now. The agentanalytics paths (`TraceManager`, `BigQueryAgentAnalyticsPlugin`, `PluginState`, `Parser`) record observational telemetry rather than persisted session-event state, so overriding them isn't needed to control ADK's generated ids/timestamps and would arguably make the telemetry less accurate. Easy to revisit if a plugin actually needs it.

DABH:
Thanks @hemasekhar-p — addressed both. I rebased onto the latest `main` and ran the full Maven build (which runs the bound `fmt:format` goal); `mvn fmt:check` now reports 0 non-complying files across all modules. I've also squashed the change into a single commit. Let me know if you still see any files being reformatted on your end.

DABH:
Also @hemasekhar-p I made a few improvements to the design/implementation based on @xumaple's feedback (updated the PR description accordingly).  Thanks again for the fast review and please just let me know what else I can do to help make this PR mergeable.  I appreciate your help and look forward to hearing from you!

DABH:
Hi @hemasekhar-p , any further thoughts here on this PR?  The equivalent Go PR got merged so just wanted to follow up.  Thanks again!

hemasekhar-p:
@DABH, thank you for the updates, To proceed with the review, could you please address merge conflicts?

DABH:
@hemasekhar-p Thanks, this has been done - can you please review?  Thanks again!

hemasekhar-p:
@DABH, thank you for addressing the feedback and resolving the merge conflicts. The PR is currently under review by our team. We will keep you updated if any further changes or any additional information is required. Thank you.

sangkyoonnam:
I need this for the same reason adk-python and adk-go added their `platform` seams: replay-safe event IDs and timestamps under a durable-execution host. adk-java is the one runtime without them. The branch conflicts with `main` again as of 4092a1fa, in two import blocks and the `Role.USER` change in `LlmEventSummarizer`, and after a local rebase with those resolved the core tests pass, so it should be a short fix on your side. Once it lands I'd like to send the Java equivalent of adk-go's `platform.WithTaskRunner` (google/adk-go#1050) as a follow-up, unless you have it in flight.

PR #1199, opened 2026-05-18
feat:contrib/model-prism: pluggable LLM backend SPI for ADK Java via `ServiceLoader`

svetanis:
## Add `model-prism`: pluggable LLM backend SPI for ADK Java via `ServiceLoader`

**Please ensure you have read the [contribution guide](./CONTRIBUTING.md) before creating a pull request.**

### Link to Issue or Description of Change

**1. Related discussion:**

- Related: #350 (OpenAI API support)
- feat: Add support for OpenAI-compatible LLM endpoints (Groq, Ollama, Azure OpenAI, etc.) #1198

**2. Description of change:**

This is a working prototype demonstrating one way to add pluggable LLM backends to ADK-Java via `ServiceLoader`.

### Problem

Python ADK gets zero-code provider switching for free — dynamic imports and LiteLLM resolve a model string from config at runtime. This PR demonstrates the Java-idiomatic equivalent: a `ServiceLoader`-based SPI where a model string in config plus a Maven dependency swap is all it takes.

### What goes where

**Into `google-adk.jar` — SPI layer (3 files):**
- `ModelProvider` — SPI interface
- `ModelProviderRegistry` — `ServiceLoader` wiring
- `OpenAiCompatibleLlm` — reusable base class for any `POST /v1/chat/completions` provider

**Into `google-adk.jar` — HTTP and serialization layer (or replaced by Google's existing implementation):**
- `OpenAiHttpClient` — HTTP client interface (default impl included; could be backed by ADK's existing `ChatCompletionsHttpClient`)
- `OpenAiMessageSerializer` — serializer interface (default impl included; could be replaced by `AdkChatSerializer` using ADK 1.1.0 internal DTOs for richer response mapping)

**Into `Runner` and `AdkWebServer.start()` (2 lines total):**
```java
ModelProviderRegistry.registerAll();
```

**Everything else stays external.** Provider JARs (Groq, Ollama, OpenRouter) live outside ADK and self-register via `META-INF/services`. No ADK PR is needed to add a new provider — dropping a JAR on the classpath in an ADK application is enough.

### What's included

| Module | Purpose |
|---|---|
| `model-prism-core` | `ModelProvider` SPI + `OpenAiCompatibleLlm` base class |
| `model-prism-groq` | Groq provider (`groq/.*`) |
| `model-prism-ollama` | Ollama provider — no API key, great for local dev |
| `model-prism-openrouter` | OpenRouter provider — free tier available |
[...]

PR #1266, opened 2026-06-13
fix: Enforce strict JSON Schema compliance to prevent OpenAI 400 Bad Request errors

### LINKED ISSUE #1265: Tool calls fail with 400 Bad Request due to strict JSON Schema violations (OpenAI)
# Tool calls fail with 400 Bad Request due to strict JSON Schema violations (OpenAI)

The ADK framework's `ChatCompletionsRequest` serialization generates JSON payloads that violate strict OpenAI/Groq requirements. Specifically, it maps empty `arguments` maps and empty function `response` objects to `null` or omits them entirely, which causes the downstream API to drop the context and hallucinate.

## 🔴 Required Information

**Describe the Bug:**
When using third-party OpenAI-compatible providers like Groq, multi-turn tool calling fails. If an agent executes a tool that requires 0 arguments (like `getCurrentTime`), the ADK serializes the assistant's message with `arguments: null` or omits the arguments field. Furthermore, during tool declaration, it omits the `parameters` schema object entirely for zero-argument functions. Strict parsers (like Groq) reject or silently drop this malformed tool history, leading the model to lose context. Consequently, the model hallucinates raw XML-like tool calls, causing a fatal 400 Bad Request error. Additionally, JSON Schema enum types for tools are incorrectly serialized in uppercase (e.g., `"STRING"` instead of `"string"`).

**Steps to Reproduce:**
1. Configure an ADK agent with multiple tools (e.g., `getCurrentTime`, `getWeather`).
2. Set the model provider to `groq/llama-3.3-70b-versatile`.
3. Send a prompt that requires executing a tool with no arguments.
4. The first tool executes, but the subsequent request to the model crashes with a 400 Bad Request.

**Expected Behavior:**
The ADK `ChatCompletionsRequest` should strictly comply with the OpenAI JSON specifications:
- `arguments` must be a non-null JSON string (e.g., `"{}"` if empty).
- `content` for tool roles must be a non-null string (e.g., `"{}"` if empty).
- `parameters` schema objects must exist in function declarations, even for zero-argument functions (e.g., `{"type":"object", "properties":{}}`).
- JSON Schema `Type` enums must be serialized in lowercase.

**Observed Behavior:**
The framework throws a 400 Bad Request exception due to model hallucinations caused by dropped context:
```text
Exception in thread "main" java.lang.RuntimeException: java.io.IOException: HTTP request failed with status: Response{protocol=h2, code=400, message=, url=https://api.groq.com/openai/v1/chat/completions} - body: {"error":{"message":"tool call validation failed: attempted to call tool 'brave_search' which was not in request.tools","type":"invalid_request_error","code":"tool_use_failed","failed_generation":"<function=brave_search>{\"query\":\"HACK stock price\"}</function>"}}
```

**Environment Details:**
 - ADK Library Version: `1.4.1-SNAPSHOT`
 - OS: Windows (also reproducible on macOS/Linux)
 - TS Version: N/A

**Model Information:**
 - Which model is being used: `groq/llama-3.3-70b-versatile`

---

## 🟡 Optional Information

 - Always (100%)

svetanis:
# Enforce strict JSON Schema compliance to prevent OpenAI 400 Bad Request errors

- Closes: #1265 (https://github.com/google/adk-java/issues/1265)
- Related: #1265

**Problem:**
When using strict OpenAI-compatible providers like Groq, empty function arguments or responses are serialized as `null` or omitted by `ChatCompletionsRequest`. Furthermore, zero-argument tools fail to declare a `parameters` schema object. This breaks strict JSON schema parsers, causing them to drop conversation history, which leads to `400 Bad Request` exceptions due to models hallucinating raw XML `<function>` tags. JSON Schema enum types are also improperly serialized in uppercase.

**Solution:**
- Added a custom `schemaNormalizerModule` to the `ObjectMapper` in `ChatCompletionsRequest.java` to force `Type` enums to serialize in lowercase (e.g., `"string"`).
- Added `enforceJsonObject` to `ChatCompletionsCommon.java` to guarantee that empty `arguments` and function `response` payloads fallback to `"{}"` instead of `null` or raw strings.
- Updated `ChatCompletionsRequest.java` to explicitly inject a `{"type":"object", "properties":{}}` parameters schema for zero-argument functions instead of omitting it.

### Testing Plan

**Unit Tests:**

_Passed: `mvn test`_
_Added `testFromLlmRequest_withEmptyFunctionArguments` to `ChatCompletionsRequestTest.java` to explicitly test `"{}"` serialization for zero-argument tools. Updated `testFromLlmRequest_withFunctionResponse` expectations._

Wired the Google native chat completion client into the external `model-prism` project ([PR #1199](https://github.com/google/adk-java/pull/1199)). Ran the integration through all the demo tests covering all main ADK features. The applications successfully execute multi-turn multi-tool conversations using strict JSON schema endpoints without throwing `400 Bad Request` due to context loss.

### Checklist

### Additional context
This change unlocks robust support for the rapidly growing ecosystem of OpenAI-compatible endpoints that enforce strict JSON Schema validation.

### DIFF
--- a/core/src/main/java/com/google/adk/models/chat/ChatCompletionsCommon.java
+++ b/core/src/main/java/com/google/adk/models/chat/ChatCompletionsCommon.java
@@ -19,8 +19,12 @@ package com.google.adk.models.chat;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.ImmutableMap;
 import com.google.genai.types.FunctionCall;
 import com.google.genai.types.Part;
 import java.util.Base64;
@@ -36,6 +40,10 @@ final class ChatCompletionsCommon {
   private ChatCompletionsCommon() {}
 
   private static final ObjectMapper objectMapper = new ObjectMapper();
+
+  static final String EMPTY_JSON_OBJECT = "{}";
+  static final ImmutableMap<String, Object> EMPTY_PARAMETERS_SCHEMA =
+      ImmutableMap.of("type", "object", "properties", ImmutableMap.of());
 
   public static final String ROLE_ASSISTANT = "assistant";
   public static final String ROLE_MODEL = "model";
@@ -157,6 +165,21 @@ final class ChatCompletionsCommon {
     }
   }
 
+  static ImmutableMap<String, Object> parseToolCallArguments(String arguments, ObjectMapper mapper)
+      throws JsonProcessingException {
+    if (arguments == null || arguments.trim().isEmpty()) {
+      return ImmutableMap.of();
+    }
+    Map<String, Object> result =
+        mapper.readValue(arguments, new TypeReference<Map<String, Object>>() {});
+    if (result == null) {
+      throw JsonMappingException.from(
+          (JsonParser) null,
+          "JSON literal 'null' is not a valid JSON object for tool call arguments");
+    }
+    return ImmutableMap.copyOf(result);
+  }
+
   /**
    * See
    * https://developers.openai.com/api/reference/resources/chat#(resource)%20chat.completions%20%3E%20(model)%20chat_completion_message_function_tool_call%20%3E%20(schema)
@@ -181,20 +204,20 @@ final class ChatCompletionsCommon {
       if (name != null) {
         fcBuilder.name(name);
       }
-      if (arguments != null && !arguments.isEmpty()) {
-        try {
-          Map<String, Object> args =
-              objectMapper.readValue(arguments, new TypeReference<Map<String, Object>>() {});
-          fcBuilder.args(args);
-        } catch (Exception e) {
-          throw new IllegalArgumentException(
-              "Failed to parse function arguments JSON: " + arguments, e);
-        }
-      }
+      fcBuilder.args(parseArguments(arguments));
       if (toolCallId != null) {
         fcBuilder.id(toolCallId);
       }
       return fcBuilder.build();
+    }
+
+    private ImmutableMap<String, Object> parseArguments(String arguments) {
+      try {
+        return parseToolCallArguments(arguments, objectMapper);
+      } catch (Exception e) {
+        throw new IllegalArgumentException(
+            "Failed to parse function arguments JSON: " + arguments, e);
+      }
     }
   }
 
--- a/core/src/main/java/com/google/adk/models/chat/ChatCompletionsRequest.java
+++ b/core/src/main/java/com/google/adk/models/chat/ChatCompletionsRequest.java
@@ -21,8 +21,12 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonValue;
+import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonSerializer;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.google.adk.JsonBaseModel;
 import com.google.adk.models.LlmRequest;
 import com.google.common.collect.ImmutableList;
@@ -32,6 +36,8 @@ import com.google.genai.types.FunctionDeclaration;
 import com.google.genai.types.FunctionResponse;
 import com.google.genai.types.GenerateContentConfig;
 import com.google.genai.types.Part;
+import com.google.genai.types.Type;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.List;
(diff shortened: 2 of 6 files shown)


hemasekhar-p:
Hi @svetanis, thank you for your contribution! We appreciate you taking the time to submit this pull request. I noticed that your changes are not covered by the current test cases and the test you included passes even without your changes. Could you please include the corresponding unit tests to verify your changes?

svetanis:
Hello @hemasekhar-p ,

Thank you for the review and the feedback! 

I've updated the PR and added three new dedicated unit tests in `ChatCompletionsRequestTest` that specifically target and verify the strict JSON Schema compliance fixes. The new tests ensure that:
1. Missing function arguments are correctly serialized as empty objects (`{}`).
2. Absent tool parameters are automatically populated with the required empty object schema.
3. Gemini's uppercase schema types (like `OBJECT` and `STRING`) are safely normalized to lowercase to satisfy strict OpenAI API validation.

I have also successfully rebased the branch on top of the latest `main` to ensure there are no conflicts with the recent `thoughtSignature` updates. Everything is folded neatly into a single commit and all tests are passing locally. 

Additionally, I ran comprehensive end-to-end testing on the rebased branch, and all of the demos (including a new one validating the `thoughtSignature` functionality) passed successfully.

Please let me know if there is anything else you need!

hemasekhar-p:
@svetanis, thank you for addressing the comments. Currently this PR is under review by our team, we will keep you posted if any additional information is required. thank you.

hemasekhar-p:
@anFatum, Could you please review this PR.

core/src/main/java/com/google/adk/models/chat/ChatCompletionsCommon.java
```diff
@@ -157,6 +161,18 @@ public Part applyThoughtSignature(Part part) {
     }
   }
 
+  /**
+   * Robust defense: Function arguments and responses MUST be JSON objects. If the LLM hallucinates
+   * "null", "NULL", "none", or conversational text, we fallback to an empty JSON object "{}". This
+   * prevents OpenAI-compatible proxies (like Groq) from throwing 400 Bad Request errors.
+   */
+  static String enforceJsonObject(String json) {
```
wikaaaaa: The raised issue and main problem if I understand correctly is that empty fields are valid (e.g. a zero-arg call legitimately has no args) but ADK serializes them incorrectly (null/omitted) which causes failures. 

`enforceJsonObject`, though, does something broader than the empty case: its `!json.trim().startsWith("{")` branch silently rewrites non-empty, non-object argument payloads (e.g. malformed/garbage input) to `{}`. That's really invalid/malformed-input handling — a separate concern from this issue. Silently dropping that content probably isn't what we want, and how invalid args should be handled deserves its own look (and a consistent approach across ADK) rather than being settled implicitly here.

And that `startsWith` branch is effectively the only thing the method does — without it, `enforceJsonObject` is just` json == null ? "{}" : json.trim()`, which is a no-op given the call sites (input is never null, and writeValueAsString/readValue already handle whitespace). So I'd suggest dropping the whole `enforceJsonObject` helper.

wikaaaaa (changes requested):
Thank you for the contribution!

We're on board with the empty/null handling: treating empty arguments/content and zero-argument parameters as valid and serializing them correctly ("{}" / {"type":"object","properties":{}}).

We have two main points we'd like to align on before approving:

- enforceJsonObject function - please see the other comment

- Inbound parsing only covers the non-streaming path. The streaming finalizer ([ChatCompletionsResponse.getFinalToolCallParts](https://github.com/google/adk-java/blob/main/core/src/main/java/com/google/adk/models/chat/ChatCompletionsResponse.java#L828)) parses tool-call args separately and isn't updated, so the same input can behave differently depending on stream=true/false. The PR should align both paths (ideally via one shared parser).

svetanis:
@wikaaaaa , 

Thank you for the feedback! Both points are addressed in the updated commit.

## 1. `enforceJsonObject` — Removed

Dropped entirely, as you suggested. You were right that its `startsWith("{")` branch was really invalid-input handling masquerading as empty-field handling, and those are separate concerns.

For legitimately absent arguments (null/empty), we now return `Map.of()` which Jackson serializes to `"{}"` naturally. For malformed LLM output (hallucinated garbage), we let the parser throw — no silent rewriting.

The one edge case worth noting: Jackson's `ObjectMapper.readValue("null", ...)` returns Java `null` rather than throwing. To keep the exception contract clean, we catch that and throw a `JsonMappingException` (a subclass of `JsonProcessingException`), so the method has a single, uniform contract — it either returns a valid `Map<String, Object>` or throws `JsonProcessingException`. Callers don't need to handle mixed exception types.

## 2. Unified Inbound Parser

Extracted a shared `parseToolCallArguments(String, ObjectMapper)` helper in `ChatCompletionsCommon` and wired it into both:

- **Non-streaming:** `ChatCompletionsCommon.Function.parseArguments()` delegates to it
- **Streaming:** `ChatCompletionsResponse.ChatCompletionChunkCollection.buildFinalResponse()` delegates to it

Both paths now behave identically for all inputs — `stream=true` and `stream=false` produce the same result for the same arguments.

## 3. Test Coverage

Added `ChatCompletionsCommonTest` with full branch coverage for the shared parser:

| Input | Expected Result | Test |
|---|---|---|
| Valid JSON (`{"key":"val"}`) | Parsed `Map` | `withValidJson` |
| `null` | `Map.of()` | `withNullString` |
| `""` | `Map.of()` | `withEmptyString` |
| `"   "` | `Map.of()` | `withWhitespaceString` |
| Invalid JSON (`"none"`, `"{bad:"`) | `JsonProcessingException` | `withInvalidJson` |
| Literal `"null"` | `JsonMappingException` | `withLiteralNullString` |

The existing `ChatCompletionsResponseTest` streaming tests (`parsesValidJsonArgs`, `handlesEmptyArgs`, `throwsOnInvalidJsonArgs`) exercise the streaming path through the same shared parser.

Rebased onto latest `main`, single clean commit, all tests passing.

wikaaaaa (approved):

core/src/main/java/com/google/adk/models/chat/ChatCompletionsCommon.java
```diff
@@ -37,6 +40,10 @@ private ChatCompletionsCommon() {}
 
   private static final ObjectMapper objectMapper = new ObjectMapper();
 
+  static final String EMPTY_JSON_OBJECT = "{}";
+  static final Map<String, Object> EMPTY_PARAMETERS_SCHEMA =
+      Map.of("type", "object", "properties", Map.of());
```
wikaaaaa: Please change `Map.of() `here and in other places to `ImmutableMap.of()` from `com.google.common.collect.ImmutableMap`;

Otherwise our internal tests fail

wikaaaaa (changes requested):

svetanis:
@wikaaaaa ,

Thank you for catching this! Completely agree — `ImmutableMap` is the right choice throughout.

Done. All `Map.of()` occurrences in the changed files have been replaced with `ImmutableMap.of()` from `com.google.common.collect.ImmutableMap`:

- `EMPTY_PARAMETERS_SCHEMA` — now `ImmutableMap.of("type", "object", "properties", ImmutableMap.of())`
- `parseToolCallArguments` — return type changed to `ImmutableMap<String, Object>`; the null/empty fast-path returns `ImmutableMap.of()`, and the Jackson-parsed result is wrapped with `ImmutableMap.copyOf(result)`

I also updated `ChatCompletionsCommonTest` to use `ImmutableMap` as the variable type in the valid-JSON test case and added an explicit `assertThat(args).isInstanceOf(ImmutableMap.class)` assertion to make the immutability contract visible in the test itself.

Rebased onto latest `main`, single clean commit, all tests passing.

wikaaaaa (approved):

PR #1268, opened 2026-06-13
fix: Prevent OkHttp idle threads from blocking JVM shutdown

### LINKED ISSUE #1267: Prevent OkHttp idle threads from blocking JVM shutdown
# Prevent OkHttp idle threads from blocking JVM shutdown

The ADK framework does not cleanly terminate the JVM upon process completion when making chat completions requests because the HTTP clients are spawning non-daemon threads with a keep-alive timeout.

## 🔴 Required Information

**Describe the Bug:**
When building an agent application using the ADK, the JVM hangs while waiting for background threads to timeout after all tasks have finished instead of cleanly exiting with code 0. This occurs because `ChatCompletionsHttpClient` and the internal `Gemini` API client construct default `OkHttpClient` instances. The default OkHttp connection pool uses a `ThreadPoolExecutor` that produces non-daemon threads. These threads prevent the JVM from shutting down until their keep-alive timer expires.

**Steps to Reproduce:**
1. Create a basic ADK agent application that makes a model request (e.g., `ToolsDemoApp`).
2. Run the application from an IDE or terminal.
3. Observe that after the final standard output logs are printed, the Java process remains active and hung before finally terminating.

**Expected Behavior:**
The JVM should exit cleanly immediately after the main thread finishes execution. The internal HTTP clients used by the ADK should be configured to use daemon threads so they do not block JVM termination.

**Observed Behavior:**
The JVM process hangs because OkHttp non-daemon threads are kept alive in the connection pool.

**Environment Details:**
 - ADK Library Version: `1.4.1-SNAPSHOT`
 - OS: Windows (also reproducible on macOS/Linux)
 - TS Version: N/A

**Model Information:**
 - Which model is being used: `groq/llama-3.3-70b-versatile` / `gemini-2.5-flash`

---

## 🟡 Optional Information

**Regression:**
No.

 - Always (100%)

svetanis:
# Prevent OkHttp idle threads from blocking JVM shutdown

- Closes: [#1267](https://github.com/google/adk-java/issues/1267)

**Problem:**
The ADK framework does not cleanly terminate the JVM upon process completion when making chat completions requests because `ChatCompletionsHttpClient` and the internal `Gemini` API client construct default `OkHttpClient` instances. These produce non-daemon threads with a keep-alive timeout, keeping the JVM alive.

**Solution:**
Created `HttpUtils.createSharedHttpClient()` utility method that returns an `OkHttpClient` using a custom dispatcher with a daemon thread factory and 5-minute timeouts. Refactored `ChatCompletionsHttpClient`, `ApiClient`, and `Gemini` to invoke this utility and maintain their own shared daemon-thread client pools.

### Testing Plan

**Unit Tests:**

_All existing core tests pass successfully. (A new test for HttpUtils can be added to verify the ThreadFactory creates daemon threads)._

Wired the Google native chat completion client into the external `model-prism` project ([PR #1199](https://github.com/google/adk-java/pull/1199)). Ran the integration through all the demo tests covering all main ADK features. Explicitly tested the native Gemini provider (via the `google_search_agent` tests) as well as third-party OpenAI-compatible providers. After the final agent response is logged, the Java process successfully exits cleanly with code 0 across all model types instead of hanging while waiting for background threads to timeout.

### Checklist

### Additional context
This fixes critical thread leaks for users running short-lived or background ADK agents.

### DIFF
--- a/core/src/main/java/com/google/adk/internal/http/HttpClientFactory.java
+++ b/core/src/main/java/com/google/adk/internal/http/HttpClientFactory.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.adk.internal.http;
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import okhttp3.Dispatcher;
+import okhttp3.OkHttpClient;
+
+/** Utility class for common HTTP client configuration across the ADK. */
+public final class HttpClientFactory {
+
+  private HttpClientFactory() {}
+
+  /**
+   * Configures a custom OkHttp dispatcher that uses daemon threads. By default, OkHttp uses
+   * non-daemon threads for its async call thread pool, which prevents the JVM from shutting down
+   * for 60 seconds (the default keep-alive) after the last streaming request completes.
+   *
+   * @param name The prefix name to use for the dispatcher threads.
+   * @return A pre-configured Dispatcher using daemon threads.
+   */
+  private static Dispatcher createDaemonDispatcher(String name) {
+    ThreadFactory daemonThreadFactory =
+        r -> {
+          Thread t = new Thread(r, name + "-Dispatcher");
+          t.setDaemon(true);
+          return t;
+        };
+    return new Dispatcher(Executors.newCachedThreadPool(daemonThreadFactory));
+  }
+
+  /**
+   * Creates a shared OkHttpClient instance equipped with a daemon thread dispatcher.
+   *
+   * @param threadName The prefix name to use for the dispatcher threads.
+   * @return A pre-configured OkHttpClient.
+   */
+  public static OkHttpClient createSharedHttpClient(String threadName) {
+    return new OkHttpClient.Builder().dispatcher(createDaemonDispatcher(threadName)).build();
+  }
+}
--- a/core/src/main/java/com/google/adk/models/Gemini.java
+++ b/core/src/main/java/com/google/adk/models/Gemini.java
@@ -19,12 +19,14 @@ package com.google.adk.models;
 import static com.google.common.base.StandardSystemProperty.JAVA_VERSION;
 
 import com.google.adk.Version;
+import com.google.adk.internal.http.HttpClientFactory;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import com.google.genai.Client;
 import com.google.genai.ResponseStream;
 import com.google.genai.types.Candidate;
+import com.google.genai.types.ClientOptions;
 import com.google.genai.types.Content;
 import com.google.genai.types.FinishReason;
 import com.google.genai.types.FunctionCall;
@@ -42,6 +44,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import okhttp3.OkHttpClient;
 import org.jspecify.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -56,6 +59,8 @@ public class Gemini extends BaseLlm {
 
   private static final Logger logger = LoggerFactory.getLogger(Gemini.class);
   private static final ImmutableMap<String, String> TRACKING_HEADERS;
+  private static final OkHttpClient SHARED_HTTP_CLIENT =
+      HttpClientFactory.createSharedHttpClient("GeminiApiClient");
 
   static {
     String frameworkLabel = "google-adk/" + Version.JAVA_ADK_VERSION;
@@ -94,6 +99,7 @@ public class Gemini extends BaseLlm {
         Client.builder()
             .apiKey(apiKey)
             .httpOptions(HttpOptions.builder().headers(TRACKING_HEADERS).build())
+            .clientOptions(ClientOptions.builder().customHttpClient(SHARED_HTTP_CLIENT).build())
             .build();
   }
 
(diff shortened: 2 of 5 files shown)


hemasekhar-p:
Hi @svetanis,  thank you for your contribution! We appreciate you taking the time to submit this pull request. Currently this PR is under review by our team, we will keep you posted if any additional information is required. thank you.

hemasekhar-p:
@anfatum, Could you please review this PR.

svetanis:
Hi @hemasekhar-p,

Thanks for the update! Just a quick heads up: I went ahead and rebased the branch onto the latest `main` so that it stays perfectly clean and conflict-free for your review. 

Looking forward to hearing back from the team!

core/src/main/java/com/google/adk/utils/HttpUtils.java
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright 2025 Google LLC
```
krwc: Nit: 2026

core/src/test/java/com/google/adk/utils/HttpUtilsTest.java
```diff
@@ -0,0 +1,29 @@
+package com.google.adk.utils;
```
krwc: Please add copyright headers to this file.

core/src/main/java/com/google/adk/utils/HttpUtils.java
```diff
@@ -0,0 +1,62 @@
...
+   *
+   * @param threadName The prefix name to use for the dispatcher threads.
+   * @return A pre-configured OkHttpClient.
+   */
+  public static OkHttpClient createSharedHttpClient(String threadName) {
+    return new OkHttpClient.Builder()
+        .dispatcher(createDaemonDispatcher(threadName))
+        .connectTimeout(Duration.ofMinutes(5))
+        .readTimeout(Duration.ofMinutes(5))
+        .writeTimeout(Duration.ofMinutes(5))
```
krwc: OkHttp's default timeouts are 10s. Any reason we decided to set them to 5mins now? If not, let's not change them to maintain backwards compat and let the callers decide.

core/src/main/java/com/google/adk/utils/HttpUtils.java
```diff
@@ -0,0 +1,62 @@
...
+package com.google.adk.utils;
+
+import java.time.Duration;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import okhttp3.Dispatcher;
+import okhttp3.OkHttpClient;
+
+/** Utility class for common HTTP client configuration across the ADK. */
+public final class HttpUtils {
```
krwc: While we can't make this class truly package private due to its use in multiple packages, I still think we should move it to somewhere else, indicating its not truly intended to be a part of public API of Java ADK.

How about we move it into `core/src/main/java/com/google/adk/internal/http`?

krwc: Nit: HttpClientFactory?

krwc (changes requested):
Thanks for the contribution! Left some comments requiring changes, but no major issues.

svetanis:
@krwc,

Thanks for the review! I've addressed all your comments:

1. **Copyright**: Updated the year to 2026 in `HttpClientFactory` and added the standard header to the test file.
2. **Rename & Move**: Moved the utility to `core/src/main/java/com/google/adk/internal/http/HttpClientFactory.java` to indicate it is not part of the public API, and renamed the test file accordingly.
3. **Timeouts**: Removed the 5-minute timeouts from `HttpClientFactory`. Let the callers (`ChatCompletionsHttpClient` and the GenAI SDK) manage their own timeouts appropriately.

Rebased onto latest `main`, single clean commit, all tests passing.

krwc (approved):
LGTM, thank you!

krwc:
FYI: we actually had to roll it back due to some issue with one of the internal systems using it. Will investigate this and try to forward fix.

svetanis:
@krwc ,

Thanks for the heads-up. Sorry to hear about the issue.

One area that might be worth investigating is the timeout configuration. The final version of the PR removed the explicit timeouts from the shared `OkHttpClient` factory, which means callers that don't set their own read/write timeouts would fall back to OkHttp's 10-second defaults. 

Happy to help with the forward fix. A couple of options:

1. **Restore defaults in `HttpClientFactory`** — add back sensible default timeouts (e.g. 5 minutes) so existing callers aren't surprised. Safest approach since we can't audit all consumers.
2. **Fix each caller** — update `ApiClient` (and any other callers) to explicitly set their own read/write timeouts. Cleaner long-term but requires knowing every consumer.

The daemon thread change itself should be safe to keep. Let me know how you'd like to proceed.

svetanis:
@krwc ,

Following up on my earlier comment. I dug into this a bit more and I think I understand what's happening. It changes both where the problem is and what I'd suggest for the fix.

**The change silently dropped Gemini/Vertex transport timeouts from infinite to 10s**, on the Gemini path.

In `google-genai:1.58.0` (`com.google.genai.ApiClient#createHttpClient`), when no custom client is set, the SDK builds its own `OkHttpClient` with `connect/read/write = 0` (infinite, deliberate for LLM streaming) and applies `httpOptions.timeout()` plus proxy. When a custom client is injected, as this PR did via `ClientOptions.customHttpClient`, it skips all of that. Our injected client set no timeouts, so it carried OkHttp's 10s defaults onto a path that was previously infinite. Any generation whose time-to-first-byte or inter-chunk gap exceeds ~10s fails with `SocketTimeoutException`. That doesn't happen on every call, just often enough. A few demo runs can miss it; production traffic won't. If you still have a trace from the rollback, I'd expect it to show a `SocketTimeoutException` on the Gemini path.

One correction to my earlier comment: I'd floated restoring ~5-minute defaults, but the right value is infinite, for exact parity with the SDK. On this path OkHttp's 10s default isn't the backwards-compatible value; the SDK's infinite is.

Proposed forward fix (keeps the JVM-shutdown fix from #1267/#1268):

- Build the injected client with `connect/read/write = 0` plus the daemon dispatcher, and make it per-`Gemini`-instance rather than a shared static. The static also capped async concurrency at OkHttp's default 5-in-flight-per-host across the whole JVM. User-supplied `Client`s stay untouched.
- Keep the daemon dispatcher for `ChatCompletionsHttpClient`; drop the `sessions/ApiClient` change (it's synchronous-only, so the daemon dispatcher was inert there).
- Add a regression test asserting the client actually handed to the SDK has infinite timeouts and daemon threads, so neither the 10s default nor the hang can silently return.

How would you like to proceed? I can open a new issue and a PR, just a PR referencing #1267/#1268, or hold off while you handle it internally. Please let me know what's easiest on your end.

krwc:
We will handle this internally, I'm working on a relevant change in the background. Thanks!

krwc:
We merged PR #1349, with a change https://github.com/google/adk-java/pull/1349/changes#diff-575f938c7b1704c98411579ce75b0fec05d612d98574c81bd93e43c80b93124bR232-R246 that should let you inject your own ExecutorService which uses daemon threads under the hood.

svetanis:
Thanks for following up, and for landing the fix. The executor hook covers our case.

PR #1273, opened 2026-06-16
fix: Enable state context in AgentExecutor

### LINKED ISSUE #1240: A2A communication drops state context by invoking 4-argument runner.runAsync
### Description
During Agent-to-Agent (A2A) communication, `AgentExecutor.java` triggers execution by calling `runner.runAsync`. However, it uses the 4-argument signature. Looking at `Runner.java`, this specific overload explicitly hardcodes `/* stateDelta= */ null`, completely dropping any custom state or metadata passed during the interaction.

To support agents that pass metadata or state deltas to guide downstream processing, `AgentExecutor ` should extract this context and pass it into the 5-argument version of `runAsync`.

### Code Reference
1. The Call Site (AgentExecutor.java)
AgentExecutor invokes the 4-argument method inside the execute pipeline:
```
return runner.runAsync(
    getUserId(ctx),
    session.id(),
    content,
    agentExecutorConfig.runConfig()); // <-- ISSUE: Invokes 4-arg variant
```
2. The Root Cause Overload (Runner.java)
In Runner.java, the 4-argument variant forces the state delta to be null:

```
public Flowable<Event> runAsync(
    String userId, String sessionId, Content newMessage, RunConfig runConfig) {
  return runAsync(userId, sessionId, newMessage, runConfig, /* stateDelta= */ null); // <-- ISSUE: Hardcoded null
}
```
### Expected Behavior
AgentExecutor should extract the incoming state/metadata from the RequestContext ctx (or the incoming Message) and invoke the 5-argument version of `runAsync` so that the `stateDelta` is properly propagated into the session history.

[Runner.java](https://github.com/user-attachments/files/28508241/Runner.java)

[AgentExecutor.java](https://github.com/user-attachments/files/28508251/AgentExecutor.java)

hemasekhar-p:

- Closes: #1240

**Problem:**
During Agent-to-Agent (A2A) communication, AgentExecutor.java triggers execution by invoking a 4-argument overload of runner.runAsync(). This specific signature in Runner.java hardcodes stateDelta to null, completely discarding any custom state or metadata passed during the interaction instead of properly propagating it to the session history

**Solution:**
To fix this issue, updated the `AgentExecutor` to extract the `metadata` from the RequestContext and directly invoking the 5-argument version of `runAsync`.

**Unit Tests:**

- [X] I have added or updated unit tests for my change.

### Checklist

### DIFF
--- a/a2a/src/main/java/com/google/adk/a2a/executor/AgentExecutor.java
+++ b/a2a/src/main/java/com/google/adk/a2a/executor/AgentExecutor.java
@@ -17,6 +17,7 @@ package com.google.adk.a2a.executor;
 
 import static java.util.Objects.requireNonNull;
 
+import com.google.adk.a2a.converters.AdkMetadataKey;
 import com.google.adk.a2a.converters.EventConverter;
 import com.google.adk.a2a.converters.PartConverter;
 import com.google.adk.agents.BaseAgent;
@@ -217,7 +218,8 @@ public class AgentExecutor implements io.a2a.server.agentexecution.AgentExecutor
                                 getUserId(ctx),
                                 session.id(),
                                 content,
-                                agentExecutorConfig.runConfig());
+                                agentExecutorConfig.runConfig(),
+                                getStateDelta(ctx));
                           });
                 })
             .concatMap(
@@ -271,6 +273,18 @@ public class AgentExecutor implements io.a2a.server.agentexecution.AgentExecutor
 
   private String getUserId(RequestContext ctx) {
     return USER_ID_PREFIX + ctx.getContextId();
+  }
+
+  private Map<String, Object> getStateDelta(RequestContext ctx) {
+    Map<String, Object> metadata = new HashMap<>();
+
+    if (ctx.getTaskId() != null) {
+      metadata.put(AdkMetadataKey.TASK_ID.getType(), ctx.getTaskId());
+    }
+    if (ctx.getContextId() != null) {
+      metadata.put(AdkMetadataKey.CONTEXT_ID.getType(), ctx.getContextId());
+    }
+    return metadata;
   }
 
   private Maybe<Session> prepareSession(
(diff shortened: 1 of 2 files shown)


hemasekhar-p:
@anFatum, Could you please review this PR.

a2a/src/main/java/com/google/adk/a2a/executor/AgentExecutor.java
```diff
@@ -273,6 +274,13 @@ private String getUserId(RequestContext ctx) {
     return USER_ID_PREFIX + ctx.getContextId();
   }
 
+  private Map<String, Object> getStateDelta(RequestContext ctx) {
+    if (ctx != null && ctx.getParams() != null) {
+      return ctx.getParams().metadata();
```
tilgalas: I think we should follow the adk python in how we handle the request context's metadata, I might be wrong here, but I think that library is more conservative in what metadata fields are included in the produced event, could you double check how the adk python solves this issue?

krwc:
Before this can go in, I dug into the Python ADK A2A executor (our reference implementation) to check the intended behavior, and I think this change needs a rethink on what gets propagated and how.

In Python ADK, the equivalent of this `execute()` step is `convert_a2a_request_to_agent_run_request()`, which builds the args for `runner.run_async()`:
https://github.com/google/adk-python/blob/main/src/google/adk/a2a/converters/request_converter.py#L100-L118

```python
custom_metadata = {}
if request.metadata:
    custom_metadata[A2A_METADATA_KEY] = request.metadata   # the caller's metadata
...
return AgentRunRequest(
    ...
    run_config=RunConfig(custom_metadata=custom_metadata),  # NOTE: no state_delta is passed here
)
```

So Python:
- propagates the caller's incoming `request.metadata`, and
- carries it via `RunConfig.custom_metadata` (which the runner then stamps onto
  each event's `custom_metadata`). 
- it leaves `state_delta` unset.

---

This PR on the other hand, puts framework-generated `taskId`/`contextId` into `stateDelta` (session state) which appears to be a completely different thing. We should align Java with Python ADK.

damianmomotgoogle:
This should be covered by #1338 and #1345

PR #1298, opened 2026-06-23
feat: Updated Spring AI to 2.0.0, ECJ, build works with Java 25

ddobrin:
* Updated Spring AI to GA version 2.0.0
*  Parent pom.xml

- Added <useIncrementalCompilation>false</useIncrementalCompilation> to the maven-compiler-plugin config (with an explanatory comment) → makes AutoValue annotation  processing deterministic across all modules and JDKs, including JDK 25
- Removed the unused ecj.version property and the org.eclipse.jdt:ecj dependencyManagement entry.

* dev/pom.xml

- Removed the unused org.eclipse.jdt:ecj dependency.

### DIFF
--- a/contrib/spring-ai/pom.xml
+++ b/contrib/spring-ai/pom.xml
@@ -29,7 +29,7 @@
     <description>Spring AI integration for the Agent Development Kit.</description>
 
     <properties>
-        <spring-ai.version>2.0.0-M3</spring-ai.version>
+        <spring-ai.version>2.0.0</spring-ai.version>
         <testcontainers.version>1.21.3</testcontainers.version>
     </properties>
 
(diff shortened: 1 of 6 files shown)


contrib/spring-ai/pom.xml
```diff
@@ -124,21 +124,21 @@
             <artifactId>spring-ai-anthropic</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
+        <!-- <dependency>
```
damianmomotgoogle: is it intentional we are commenting out those dependencies?

ddobrin: These are deprecated libraries. Vertex SDK is at the end of its lifecycle at the end of June.
Will remove from pom.xml

damianmomotgoogle (approved):

damianmomotgoogle (changes requested):
commit message still needs updating with "feat:" - otherwise it will be dropped from release notes

damianmomotgoogle (approved):

PR #1318, opened 2026-07-01
BQAA (Java): preview-readiness fixes — redaction, table bootstrap, drop stats, reliability

### LINKED ISSUE #1316: BQAA Java preview readiness tracker: correctness, lifecycle, and Python contract parity
## Summary

This is the consolidated readiness tracker for the Java BigQuery Agent Analytics (BQAA) plugin.

Current verdict: **not ready to exit Preview or claim Python-compatible Preview behavior**. Full Python feature parity is not required, but the P1 items below can produce incorrect analytics, duplicate telemetry, retained per-invocation resources, or incomplete cleanup on supported Java runtime paths.

Latest comparison, pinned on 2026-07-10:

- Java `google/adk-java@08f4fdc6c9ea80ec46723b2c6b71a4409c77feed`
- Python `google/adk-python@9d306f5d322eccbc5022ac72407fa8a994f64cf6`

## P1: must fix before broad/public Preview

- [ ] **Make trace state safe for concurrent `ParallelAgent` branches.**
  - Java stores one `TraceManager` per invocation and one global LIFO stack inside it.
  - `ParallelAgent` merges concurrently scheduled branches with the same invocation ID. If branch A starts, branch B starts, and A completes first, A pops B's span.
  - Use task-local, branch-keyed, or otherwise ownership-checked trace state.
  - Add a deterministic same-invocation overlap test that completes branches out of start order and validates span IDs, parents, and STARTING/COMPLETED pairing.

- [ ] **Stop creating plugin-owned OpenTelemetry spans.**
  - `TraceManager.pushSpan()` starts a real OTel span and pop/cleanup ends it.
  - With an SDK exporter configured, BQAA therefore exports a duplicate analytics span tree alongside ADK's framework spans.
  - Match Python's ID-only internal execution tree while inheriting an ambient `trace_id` when available.
  - Add an in-memory exporter regression test asserting BQAA exports zero plugin-owned spans.

- [ ] **Apply sensitive-key redaction at the final attributes boundary.**
  - Content and session-metadata paths now use the redacting formatter, but the final assembled attributes map is not passed through that boundary.
  - `state_delta`, custom tags, labels, and other extra attributes can therefore carry sensitive-key values into BigQuery unchanged.
  - Redact the complete attributes tree immediately before serialization, regardless of which producer populated it. Python currently has the same gap, so parity is not a safety defense.
  - Add nested secret-key fixtures for state deltas, custom tags, labels, and generic metadata.

- [ ] **Fix per-invocation processor/task lifecycle.**
  - `BatchProcessor.start()` discards the `ScheduledFuture` returned by `scheduleWithFixedDelay()`.
  - Per-invocation `close()` does not cancel it, so every completed invocation leaves a periodic task retaining its closed processor/writer until plugin-wide shutdown.
  - Store/cancel the task during idempotent close and add a high-cardinality invocation test proving scheduled work/resources are released.

- [ ] **Make append, final drain, and timeout cleanup genuinely bounded.**
  - `writer.append(...).get()` has no total deadline, and final processor drain runs outside the effective `shutdownTimeout` bound.
  - Pending parse/offload futures are not cancelled or lifecycle-gated after timeout. A late completion can recreate a `BatchProcessor` for an invocation that has already been finalized, with no later invocation cleanup.
  - Apply a total append/drain deadline, reject/account appends after invocation close, and test a never-completing append plus a late GCS/parse completion.

- [ ] **Surface and account writer-construction and continuation failures.**
  - `StreamWriter` creation occurs inside `computeIfAbsent`; if the future completes exceptionally, the pending task is removed without a durable error signal or drop counter.
  - Rows can therefore disappear before a processor exists, outside the current loss-accounting path.
  - Record a specific drop reason, emit an observable error, and make late/exceptional continuations respect the invocation lifecycle gate.

- [ ] **Correct HITL and long-running-tool pause/resume semantics.**
  - Java currently labels both HITL `FunctionCall` requests and `FunctionResponse` results as `*_COMPLETED`.
  - User-message handling inspects `FunctionCall` where resumed input arrives as `FunctionResponse`.
  - Emit pairable `TOOL_PAUSED` / resumed `TOOL_COMPLETED` rows with `function_call_id` and `pause_kind`, without double-routing HITL completions.
  - Add a Runner-level two-turn test for HITL and non-HITL long-running tools.

- [ ] **Add terminal agent/run failure analytics and guaranteed cleanup.**
  - Java's plugin API exposes model/tool error callbacks but no agent/run error callbacks.
  - `afterAgent` / `afterRun` are concatenated after successful completion, so an upstream exception can skip `AGENT_ERROR`, `INVOCATION_ERROR`, per-invocation flush, and trace cleanup.
  - Add framework error hooks and finally-style invocation cleanup that preserves the original exception.

## Contract gaps: fix or explicitly exclude from a narrow Preview

- [ ] **Align or version the `AGENT_RESPONSE` row/view contract.**
  - Java emits `content.text_summary` and exposes `text_summary` in its view.
  - Python emits `content.response` and exposes `response_text`.
  - The same event type should be queryable across SDKs; add a cross-language golden-row test.

- [ ] **Complete supported workflow taxonomy.**
  - Java runtime models transfer, compaction, and long-running tool IDs, but Java BQAA does not emit Python-equivalent `AGENT_TRANSFER`, `EVENT_COMPACTION`, and pairable `TOOL_PAUSED` rows/views.
  - Implement these before a parity claim, or list each omission in Preview scope documentation.

- [ ] **Add the minimum `attributes.adk` producer envelope.**
  - At minimum stamp schema version, app name, and available source-event identity/branch fields.
  - Python-only node/scope fields can follow when Java exposes equivalent runtime data.

- [ ] **Publish an executable Java Preview guide and data contract.**
  - Cover dependency/setup, ADC and IAM, dataset/table/view defaults, explicit close behavior, drop-stat monitoring, GCS/privacy controls, event taxonomy, row/view contract, and documented Python-parity exclusions.

## P2 reliability and operational readiness

- [ ] Move first-use table bootstrap off the agent callback thread; coalesce concurrent setup and use bounded retry/backoff after failures.
- [ ] Make `enabled=false` avoid ADC resolution and BigQuery client initialization.
- [ ] Add explicit loss accounting for setup-not-ready, writer construction, parse timeout, and rejected late continuations.
- [ ] Add a gated real-BigQuery smoke test using public construction, real Storage Write append/drain, and query readback. The current `BigQueryAgentAnalyticsPluginE2ETest` mocks BigQuery, `BigQueryWriteClient`, and `StreamWriter`.
- [ ] Decide and document Java/Python default differences (`agent_analytics.events`, views disabled vs Python `agent_events`, views enabled).
- [ ] Make model, tool, and BigQuery exception-message capture null-safe; Java exceptions may legally return a null message, while the generated event value objects require a non-null optional payload.
- [ ] Validate all duration and retry settings at config build time, including positive flush/shutdown durations and valid retry-attempt bounds.

## Lower-priority Python gaps that do not independently block a narrow Preview

- [ ] Schema-first payload-column projection / denylist.
- [ ] Generic custom-metadata allowlisting.
- [ ] Full node/scope/route/render/rewind envelope parity.
- [ ] Structured LLM tool declaration parity.
- [ ] Public flush/view-management helper parity.

These should remain roadmap items unless the Preview explicitly promises full Python query/config parity.

## Confirmed fixed on current Java `main`

- [x] Sensitive-key redaction for content/session-metadata paths and fail-closed custom content formatting.
  - Final assembled attributes remain open under the P1 above and are not included in this completed item.
- [x] Empty state deltas no longer emit `STATE_DELTA`.
- [x] Timestamp partitioning for newly created tables.
- [x] Failed table ensure is retryable; schema drift is logged.
- [x] Root agent initialization and `Event.author` fallback.
- [x] Drop-stat folding, regional routing, shutdown hook, headers, and numeric/config validation.

## Acceptance tests for Preview readiness

- [ ] Same-invocation `ParallelAgent` overlap preserves branch-local span ownership.
- [ ] Recording OTel exporter receives zero BQAA-owned spans.
- [ ] Final attributes redaction covers nested secrets in state deltas, custom tags, labels, and generic metadata.
- [ ] Many completed invocations leave no periodic flush tasks/processors/writers retained.
- [ ] `shutdownTimeout` bounds a never-completing append and final drain.
- [ ] A parse/offload future completing after timeout cannot recreate an invocation processor.
- [ ] Writer-construction failures increment a specific drop counter and emit an observable error.
- [ ] HITL and ordinary long-running tools produce correctly named and pairable pause/resume rows.
- [ ] Agent and invocation exceptions produce terminal rows, preserve the original error, and finalize state exactly once.
- [ ] Null-message model/tool/BigQuery exceptions do not break analytics callbacks.
- [ ] Invalid flush/shutdown durations and retry settings fail at config construction.
- [ ] Java/Python golden fixtures agree on common event names, row fields, content JSON, trace linkage, and view projections.
- [ ] Slow/failing first-use table setup does not stall agent progress or trigger one control-plane RPC per event.
- [ ] Real BigQuery write/readback smoke test passes in a disposable dataset.

## Current verification signal

- Java focused BQAA suite: **135 passed, 0 failed**.
- Java full core main profile: **1,591 passed, 0 failed/errors, 24 skipped**; additional profiles also passed.
- Python pinned plugin suite: **306 passed, 6 skipped**.
- These suites do not currently cover the production shapes listed in the acceptance checklist above.

## Pinned references

- Java plugin package: https://github.com/google/adk-java/tree/08f4fdc6c9ea80ec46723b2c6b71a4409c77feed/core/src/main/java/com/google/adk/plugins/agentanalytics
- Java `ParallelAgent`: https://github.com/google/adk-java/blob/08f4fdc6c9ea80ec46723b2c6b71a4409c77feed/core/src/main/java/com/google/adk/agents/ParallelAgent.java#L172-L184
- Python reference plugin: https://github.com/google/adk-python/blob/9d306f5d322eccbc5022ac72407fa8a994f64cf6/src/google/adk/plugins/bigquery_agent_analytics_plugin.py

Historical investigation and withdrawn/down-scoped findings remain available in the comments below.

caohy1988:
## Summary

Addresses preview-readiness gaps in the Java BigQuery Agent Analytics (BQAA) plugin tracked in google/adk-java#1316, bringing it closer to parity with the Python plugin and hardening the write path.

> [!IMPORTANT]
> This change was **not built or tested locally** (the full Maven build was not run in the authoring environment). Please rely on CI and review. Unit tests for the new behaviors still need to be added — see "Tests to add" below. Findings were verified by direct source review against the issue-pinned SHA `ce6af87`, which is byte-identical to current `main` for this package.

Fixes #1316 (partially — remaining deferred items listed below).

## P1 — preview blockers

- **Secret redaction** (`JsonFormatter`): sensitive keys (`client_secret`, `access_token`, `refresh_token`, `id_token`, `api_key`, `password`, and any `temp:*` key) are now redacted to `[REDACTED]` during truncation, matching the Python plugin's `_SENSITIVE_KEYS`. Case-insensitive; redaction does not set the `is_truncated` flag (parity). This covers tool args/results, session state, usage metadata, and custom-tag paths that flow through `smartTruncate`.
- **Empty `STATE_DELTA` guard** (`onEventCallback`): a `STATE_DELTA` row is only emitted when `event.actions().stateDelta()` is non-empty, matching Python (no row for empty deltas).
- **Table bootstrap made production-safe** (`ensureTableExists` / `ensureTableExistsOnce`):
  - New tables are created with **DAY time-partitioning on `timestamp`** (parity with Python; enables partition pruning), in addition to the existing clustering.
  - `tableEnsured` is now set **only after a successful setup**, so a transient first-run failure (auth blip, missing dataset, quota) is retried on later events instead of permanently disabling table create/upgrade for the plugin instance.
  - Concurrent-create (`already exists`) is treated as success rather than a hard failure.
- **`root_agent_name` is populated** (`TraceManager.initTraceIfNeeded`, called from `getAttributes`): `attributes.root_agent_name` now resolves to the real root agent name instead of the sentinel default. Null-safe.
- **No-current-agent guard** (`resolveAgentName`): the `agent` column is resolved defensively; workflow-driven callbacks with no current agent fall back to `"unknown"` instead of NPE-dropping the row.

## P2 — reliability hardening

- **Drop accounting** (`BatchProcessor` + `PluginState` + `getDropStats()`): rows lost to a full queue, append errors, or serialization errors are now counted and exposed via `plugin.getDropStats()` (`queue_full` / `append_error` / `serialization_error`). Counters survive per-invocation processor churn by folding into plugin-level totals at close.
- **Storage Write regional routing**: the `StreamWriter` is now built with `.setLocation(config.location())` so append RPCs route to the dataset's region (previously only the control-plane client set location; non-US/EU appends risked stream-not-found).
- **JVM shutdown hook**: public construction paths register a best-effort drain (`state.close()` bounded by `shutdownTimeout`) so a host that never calls `close()` still flushes queued rows at exit. The package-private test constructor does **not** register a hook (avoids hook leakage in tests).
- **Schema auto-upgrade always diffs** (`maybeUpgradeSchema`): removed the version-label-only early return that Python deliberately avoids — a table stamped with the current label but missing columns is now reconciled. Incompatible non-STRUCT type drift is logged (was silently ignored, later surfacing as opaque append failures).

## P3 — operational polish

- **Config validation**: `batchSize`, `queueMaxSize`, and `maxContentLength` are validated as positive in `build()`.
- **View DDL identifier validation**: `createAnalyticsViews` refuses to interpolate a project/dataset/table/viewPrefix that isn't a safe identifier (guards the string-formatted DDL).
- **Recursion depth guard** in `recursiveSmartTruncate` (prevents `StackOverflowError` on pathological nesting) + a log line on the string-fallback path.
- **Hot-path INFO logs lowered to FINE** (`isProcessed`, pending-task wait/removal, stale-span clearing, table-exists).
- **`EventData.java`**: added the missing Apache license header and fixed stale `_log_event` Javadoc.

## Deferred (intentionally not in this PR)

These need behavior/parity decisions and tests I did not want to ship blind:

- **Preserve the BQAA internal span tree under ambient OpenTelemetry** (#1316 P1 item): still prefers ambient span IDs. Changing this alters correlation semantics and needs tests.
- **Full ADK `attributes.adk` envelope parity** (#1316 P2): route/render/rewind metadata, source-event node/branch/scope, long-running pair keys.
- **Long-running (non-HITL) tool resume → `TOOL_COMPLETED` with `pause_kind`/`function_call_id` pair keys** (#1316 P1).
- **Fully off-thread first-event table ensure**: `ensureTableExists` is still synchronous on the first event's callback. Moving it off-thread safely needs append-gating so early rows aren't lost; deferred to avoid an untested concurrency change. (Note for whoever picks this up: snapshot any mutable `InvocationContext`/`Session` data before crossing threads.)
- **Default table/view naming decision** (`agent_analytics.events` + `createViews=false` vs Python `agent_events` + `create_views=True`): a product/docs decision, not a code fix.

Plugin-owned OTel span export (#1316 comment) is related to the span-tree item and is deferred with it.

## Tests to add

- Redaction across nested maps, tool args/results, session state, custom tags; `is_truncated` unaffected by redaction.
- Empty `state_delta` produces no `STATE_DELTA` row.
- Table creation sets DAY partitioning + clustering; failed first-run setup is retried on later events.
- `getDropStats()` increments on queue-full / append-error / serialization-error paths.
- Schema upgrade reconciles a same-label table that is missing a column.
- Config builder rejects non-positive `batchSize`/`queueMaxSize`/`maxContentLength`.
- Shutdown hook drains queued rows when `close()` is not called explicitly.

### DIFF
--- a/core/src/main/java/com/google/adk/plugins/agentanalytics/BatchProcessor.java
+++ b/core/src/main/java/com/google/adk/plugins/agentanalytics/BatchProcessor.java
@@ -28,6 +28,7 @@ import com.google.cloud.bigquery.storage.v1.AppendRowsResponse;
 import com.google.cloud.bigquery.storage.v1.Exceptions.AppendSerializationError;
 import com.google.cloud.bigquery.storage.v1.StreamWriter;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
@@ -37,6 +38,7 @@ import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.apache.arrow.memory.BufferAllocator;
@@ -67,6 +69,11 @@ class BatchProcessor implements AutoCloseable {
   final AtomicBoolean flushLock = new AtomicBoolean(false);
   private final Schema arrowSchema;
   private final VectorSchemaRoot root;
+
+  // Drop accounting so hosts can programmatically detect lost analytics rows.
+  private final AtomicLong droppedQueueFull = new AtomicLong();
+  private final AtomicLong droppedAppendError = new AtomicLong();
+  private final AtomicLong droppedSerializationError = new AtomicLong();
 
   public BatchProcessor(
       StreamWriter writer,
@@ -105,6 +112,7 @@ class BatchProcessor implements AutoCloseable {
 
   public void append(Map<String, Object> row) {
     if (!queue.offer(row)) {
+      droppedQueueFull.incrementAndGet();
       logger.warning("BigQuery event queue is full, dropping event.");
       return;
     }
@@ -139,6 +147,7 @@ class BatchProcessor implements AutoCloseable {
         try (ArrowRecordBatch recordBatch = new VectorUnloader(root).getRecordBatch()) {
           AppendRowsResponse result = writer.append(recordBatch).get();
           if (result.hasError()) {
+            droppedAppendError.addAndGet(batch.size());
             logger.severe("BigQuery append error: " + result.getError().getMessage());
             for (var error : result.getRowErrorsList()) {
               logger.severe(
@@ -153,6 +162,7 @@ class BatchProcessor implements AutoCloseable {
           Thread.currentThread().interrupt();
         }
         if (e.getCause() instanceof AppendSerializationError ase) {
+          droppedSerializationError.addAndGet(batch.size());
           logger.log(
               Level.SEVERE, "Failed to write batch to BigQuery due to serialization error", ase);
           Map<Integer, String> rowIndexToErrorMessage = ase.getRowIndexToErrorMessage();
@@ -167,6 +177,7 @@ class BatchProcessor implements AutoCloseable {
                 "AppendSerializationError occurred, but no row-specific errors were provided.");
           }
         } else {
+          droppedAppendError.addAndGet(batch.size());
           logger.log(Level.SEVERE, "Failed to write batch to BigQuery", e);
         }
       } finally {
@@ -245,6 +256,17 @@ class BatchProcessor implements AutoCloseable {
         }
       }
     }
+  }
+
+  /**
+   * Returns a snapshot of dropped-row counters keyed by reason ({@code queue_full}, {@code
+   * append_error}, {@code serialization_error}). Non-zero values indicate lost analytics rows.
+   */
+  ImmutableMap<String, Long> getDropStats() {
+    return ImmutableMap.of(
+        "queue_full", droppedQueueFull.get(),
+        "append_error", droppedAppendError.get(),
+        "serialization_error", droppedSerializationError.get());
   }
 
   @Override
--- a/core/src/main/java/com/google/adk/plugins/agentanalytics/BigQueryAgentAnalyticsPlugin.java
+++ b/core/src/main/java/com/google/adk/plugins/agentanalytics/BigQueryAgentAnalyticsPlugin.java
@@ -215,10 +215,14 @@ public class BigQueryAgentAnalyticsPlugin extends BasePlugin {
             throw e;
           }
         }
+        tableReady = true;
       } else if (config.autoSchemaUpgrade()) {
-        maybeUpgradeSchema(bigQuery, table);
+        // Only treat the table as ready if the schema upgrade actually succeeded, so a failed
+        // upgrade is retried on a later event instead of being masked by tableEnsured=true.
+        tableReady = maybeUpgradeSchema(bigQuery, table);
+      } else {
+        tableReady = true;
       }
-      tableReady = true;
     } catch (BigQueryException e) {
       processBigQueryException(e, "Failed to check or create/upgrade BigQuery table: " + tableId);
     } catch (RuntimeException e) {
@@ -289,7 +293,7 @@ public class BigQueryAgentAnalyticsPlugin extends BasePlugin {
     Map<String, Object> row = new HashMap<>();
     row.put("timestamp", Instant.now());
     row.put("event_type", eventType);
-    row.put("agent", resolveAgentName(invocationContext));
+    row.put("agent", resolveAgentName(invocationContext, eventData));
     row.put("session_id", invocationContext.session().id());
     row.put("invocation_id", invocationContext.invocationId());
     row.put("user_id", invocationContext.userId());
@@ -352,18 +356,28 @@ public class BigQueryAgentAnalyticsPlugin extends BasePlugin {
 
   /**
    * Resolves the agent name defensively. Workflow-driven callbacks may have no current agent; fall
-   * back to a sentinel rather than letting an NPE drop the row.
+   * back to the event author (via {@code EventData.fallbackAgentName}, mirroring the Python plugin)
+   * and finally to a sentinel, rather than letting an NPE drop the row.
    */
-  private static String resolveAgentName(InvocationContext invocationContext) {
+  private static String resolveAgentName(
+      InvocationContext invocationContext, Optional<EventData> eventData) {
     try {
       BaseAgent agent = invocationContext.agent();
       if (agent != null && agent.name() != null) {
         return agent.name();
       }
     } catch (RuntimeException e) {
-      // Fall through to the sentinel below.
+      // Fall through to the author/sentinel fallback below.
     }
-    return "unknown";
+    return eventData.flatMap(EventData::fallbackAgentName).orElse("unknown");
+  }
+
+  private static EventData.Builder withFallbackAgent(
+      EventData.Builder builder, @Nullable String author) {
+    if (author != null && !author.isEmpty()) {
+      builder.setFallbackAgentName(author);
+    }
+    return builder;
   }
 
   private ResolvedTraceIds getResolvedTraceIds(
@@ -525,12 +539,14 @@ public class BigQueryAgentAnalyticsPlugin extends BasePlugin {
     Completable logCompletable = Completable.complete();
     if (!event.actions().stateDelta().isEmpty()) {
       EventData.Builder eventDataBuilder =
-          EventData.builder()
-              .setExtraAttributes(
-                  ImmutableMap.<String, Object>builder()
-                      .put("state_delta", event.actions().stateDelta())
-                      .put("author", event.author())
-                      .buildOrThrow());
+          withFallbackAgent(
+              EventData.builder()
+                  .setExtraAttributes(
+                      ImmutableMap.<String, Object>builder()
+                          .put("state_delta", event.actions().stateDelta())
+                          .put("author", event.author())
+                          .buildOrThrow()),
+              event.author());
       logCompletable =
           logEvent(
               "STATE_DELTA",
@@ -618,7 +634,11 @@ public class BigQueryAgentAnalyticsPlugin extends BasePlugin {
                     invocationContext,
                     contentObject,
                     a2aTruncated.isTruncated() || contentTruncated,
-                    Optional.of(EventData.builder().setExtraAttributes(extraAttributes).build())));
+                    Optional.of(
+                        withFallbackAgent(
+                                EventData.builder().setExtraAttributes(extraAttributes),
+                                event.author())
+                            .build())));
       }
     }
 
@@ -653,7 +673,11 @@ public class BigQueryAgentAnalyticsPlugin extends BasePlugin {
                     invocationContext,
                     visibleContent,
                     false,
-                    Optional.of(EventData.builder().setExtraAttributes(extraAttributes).build())));
+                    Optional.of(
+                        withFallbackAgent(
+                                EventData.builder().setExtraAttributes(extraAttributes),
+                                event.author())
+                            .build())));
       }
     }
 
--- a/core/src/main/java/com/google/adk/plugins/agentanalytics/BigQueryAgentAnalyticsPlugin.java
+++ b/core/src/main/java/com/google/adk/plugins/agentanalytics/BigQueryAgentAnalyticsPlugin.java
@@ -51,6 +51,7 @@ import com.google.cloud.bigquery.StandardTableDefinition;
 import com.google.cloud.bigquery.Table;
 import com.google.cloud.bigquery.TableId;
 import com.google.cloud.bigquery.TableInfo;
+import com.google.cloud.bigquery.TimePartitioning;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
@@ -65,10 +66,12 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.jspecify.annotations.Nullable;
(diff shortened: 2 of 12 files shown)


caohy1988:
## Review against #1316 preview-readiness issue

I re-reviewed this PR against the tracked preview blockers in #1316. This is a strong partial fix, but I would keep it draft for now. A few issue-tracked blockers remain, and I found two regressions/holes in the new implementation that should be fixed before marking ready.

### Findings

1. **P1: table bootstrap still stops retrying after schema-upgrade failure.**
   `ensureTableExists()` marks the table ready after `maybeUpgradeSchema()` returns (`BigQueryAgentAnalyticsPlugin.java:218-221`), but `maybeUpgradeSchema()` catches and logs `BigQueryException` without surfacing failure (`BigQueryUtils.java:269-287`). Result: a failed upgrade can still set `tableEnsured=true` (`BigQueryAgentAnalyticsPlugin.java:163-172`), leaving the instance permanently “ensured” while appends may fail on missing columns. Please make schema upgrade return success/failure or rethrow, and only set `tableEnsured` after create/existing-table/schema-upgrade all succeed.

2. **P1: license-header cleanup is still incomplete.**
   `EventData.java` is fixed, but `PluginState.java` still starts at `package ...` with no Apache header (`PluginState.java:1`). Since #1316 tracks this as release/preview hygiene, this should be fixed before preview.

3. **P2: drop stats miss close/shutdown-time drops.**
   `PluginState.ensureInvocationCompleted()` folds counters before `processor.close()` (`PluginState.java:266-270`), and global `close()` does the same (`PluginState.java:329-331`). But `BatchProcessor.close()` can still flush queued rows and increment drop counters (`BatchProcessor.java:273-276`, write failures counted at `BatchProcessor.java:149-181`). Drops that happen during final drain are then removed with the processor and never reach `getDropStats()`. Fold after final close/drain, or fold both before and after.

4. **P2: schema drift logging still ignores mode drift.**
   The PR logs incompatible type drift only (`BigQueryUtils.java:328-342`). The #1316 follow-up finding was type/mode drift; mode differences remain silent because `schemaFieldsMatch()` does not compare `Field.Mode` for top-level or nested fields (`BigQueryUtils.java:299-345`). Add mode checks and warnings for nested non-STRUCT drift too.

5. **P2: no-current-agent fallback is only partial.**
   The NPE/drop is fixed, but the `agent` column now becomes `"unknown"` (`BigQueryAgentAnalyticsPlugin.java:292`, `357-366`) even when `Event.author()` is available on source-event rows. The issue asked for Python-style fallback to `Event.author`; Java currently only stores that as `attributes.source_event_author` (`BigQueryAgentAnalyticsPlugin.java:644-656`).

### Coverage against #1316

Good fixes landed: redaction, empty `STATE_DELTA` guard, timestamp partitioning on create, retry after most table setup failures, root agent initialization, defensive agent lookup, regional `StreamWriter.setLocation`, config validation, view identifier validation, `EventData` header/Javadoc, recursion depth guard, and quieter logs.

Still intentionally unresolved from the issue: HITL/user-message resume parity, added Python event types and views, ambient OTel/internal span-tree semantics, plugin-owned OTel span export, full `attributes.adk` envelope, off-thread first-event ensure, default table/view naming decision, per-invocation writer scaling, and salvage/DLQ/retry behavior for bad batches.

### Readiness

Recommendation: keep this PR as draft until the P1s above are fixed and tests are added. Current GH status also has `cla/google` failing; only lightweight checks passed so far.

caohy1988:
Thanks — all five findings validated against source and fixed in 11d2fab. Point by point:

1. **Table bootstrap retry after upgrade failure — fixed.** `maybeUpgradeSchema()` now returns `boolean` (true = up-to-date or upgraded successfully; false = update threw). `ensureTableExists()` sets `tableReady` from that result on the existing-table path, so a failed schema upgrade no longer flips `tableEnsured=true` and is retried on the next event. Create and no-upgrade paths set `tableReady=true` explicitly.

2. **License header — fixed.** Added the Apache header to `PluginState.java`. Swept the rest of the package; it was the only remaining file starting at `package …`.

3. **Drop stats miss final-drain drops — fixed.** `foldDropStats(processor)` now runs **after** `processor.close()` in both `ensureInvocationCompleted()` and `close()`, so rows dropped during the closing flush are counted before the processor is discarded. (Counters are `AtomicLong`s on the processor and are still readable post-close.)

4. **Mode drift — fixed.** `schemaFieldsMatch()` now compares `Field.Mode` in addition to `StandardType`, with null normalized to `NULLABLE`. The warning fires on type **or** mode drift and covers nested non-STRUCT fields (same method recurses into STRUCTs). Message now reports `type/mode` on both sides.

5. **Agent fallback to `Event.author` — implemented.** Added `EventData.fallbackAgentName` (Optional). `resolveAgentName()` now returns current agent → `fallbackAgentName` → `"unknown"`. The source-event rows in `onEventCallback` (`STATE_DELTA`, `A2A_INTERACTION`, `AGENT_RESPONSE`) now pass `event.author()` as the fallback, matching the Python behavior. HITL rows keep the live agent (an agent is always present mid-invocation there), so they were left as-is; let me know if you'd rather thread author there too.

Everything under "still intentionally unresolved" in your review remains deferred (span-tree/OTel semantics, ADK envelope, HITL long-running resume + pair keys, off-thread first-event ensure, default naming, per-invocation writer scaling, DLQ/salvage). Keeping the PR **draft**: still not built locally and unit tests for these behaviors are pending. The `cla/google` check is the CLA signature, separate from the code.

caohy1988:
Two follow-ups in 08fe970:

**1. Internal span tree under ambient OTel (previously deferred #1316 P1).** `getResolvedTraceIds` now always sources `span_id`/`parent_span_id` from the BQAA internal execution tree (`getCurrentSpanAndParent`) instead of preferring ambient OpenTelemetry span IDs, so `parent_span_id` always references another logged row rather than an unlogged framework span. Ambient OTel still governs `trace_id` (via `getTraceId`). The LLM/tool completed+error callbacks and `getCompletedEventData` now unconditionally record the popped internal span (dropped the `!hasAmbient` gating); the now-dead `getAmbientSpanAndParent()` helper is removed.

Note: this does **not** stop the plugin from creating real OTel spans — the duplicate-export concern is separate and still deferred. This change only fixes which span IDs land in the rows.

**2. Unit tests** (JUnit4/vintage, matching existing style):
- `JsonFormatterTest`: sensitive-key redaction — top-level, case-insensitive, `temp:` prefix, nested; asserts redaction does not set `is_truncated` — plus the recursion depth guard.
- `BigQueryLoggerConfigTest` (new): `build()` rejects non-positive `batchSize`/`queueMaxSize`/`maxContentLength`.
- `TraceManagerTest`: `initTraceIfNeeded` sets the root agent name from the context, and a null agent keeps the sentinel.

Still not executed locally (no Maven build in my environment), so CI is the gate. Remaining deferred items from your review are unchanged: ADK envelope, HITL long-running resume + pair keys, plugin-owned OTel span *export*, off-thread first-event ensure, default naming, per-invocation writer scaling, DLQ/salvage.

hemasekhar-p:
Hi @caohy1988, thank you for your contribution! We appreciate you taking the time to submit this pull request. I noticed that the pre-checks are currently failing and make sure your PR consist of single commit as per Contribution Policy. Could you please take a look and address those comments?

PR #1331, opened 2026-07-06
fix(skills): prevent path traversal in LocalSkillSource

adilburaksen:
### What
Prevents path traversal in `LocalSkillSource`. Skill names and resource paths from caller input were resolved directly against `skillsBasePath` without validating containment, so a name/path containing `..` or an absolute path could escape the skills base directory (arbitrary read of files outside the configured skills root, e.g. credentials).

Adds `validatePathWithinBase`, which rejects absolute components and any component that normalizes outside its base directory, and routes `listResources`, `findResourcePath` and `findSkillMdPath` through it. The helper returns the validated, normalized path so callers reuse it instead of re-resolving `skillsBasePath.resolve(skillName)`. Component parsing uses `base.getFileSystem()` so it matches the filesystem `base.resolve()` uses when `skillsBasePath` comes from a non-default provider, and the error code is passed in so resource-path checks report `RESOURCE_NOT_FOUND` while skill-name checks report `SKILL_NOT_FOUND`.

### Note
This replaces #1228. That PR's branch was accidentally force-pushed on my side, which left #1228 in an un-reopenable state. This PR is the same branch with all review feedback from @wikaaaaa already applied:
- Warning/message and error-code handling addressed; the traversal comment no longer references the unused `Path.of()`.
- All five traversal tests now assert the specific `SkillSourceException` error code in addition to the message.

Single commit, CI-clean, scoped to `LocalSkillSource` + its test.

### Testing
`mvn -pl core test -Dtest=LocalSkillSourceTest` passes locally (21 tests, 0 failures).

### DIFF
--- a/core/src/main/java/com/google/adk/skills/LocalSkillSource.java
+++ b/core/src/main/java/com/google/adk/skills/LocalSkillSource.java
@@ -45,12 +45,18 @@ public final class LocalSkillSource extends AbstractSkillSource<Path> {
 
   @Override
   public Single<ImmutableList<String>> listResources(String skillName, String resourceDirectory) {
-    Path skillDir = skillsBasePath.resolve(skillName);
+    Path skillDir;
+    Path resourceDir;
+    try {
+      skillDir = validatePathWithinBase(skillsBasePath, skillName, SKILL_NOT_FOUND);
+      resourceDir = validatePathWithinBase(skillDir, resourceDirectory, RESOURCE_NOT_FOUND);
+    } catch (SkillSourceException e) {
+      return Single.error(e);
+    }
     if (!isDirectory(skillDir)) {
       return Single.error(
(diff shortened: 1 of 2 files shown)


adilburaksen:
@wikaaaaa this is the follow-up to #1228 — its branch got accidentally force-pushed on my end and GitHub wouldn't let me reopen it, so I re-opened here from the same branch. Both of your remaining comments are addressed: the traversal comment no longer mentions the unused `Path.of()`, and all five traversal tests now assert the specific error code (`SKILL_NOT_FOUND` / `RESOURCE_NOT_FOUND`) in addition to the message. Single commit, `mvn -pl core test -Dtest=LocalSkillSourceTest` green. Sorry for the churn — happy to adjust anything.

wikaaaaa (approved):

PR #1358, opened 2026-07-18
Fix: live connection drops all but the last parallel function call

### LINKED ISSUE #1357: Live (BIDI) connection drops all but the last parallel function call
When a Gemini live session returns multiple function calls in a single `toolCall`
message, ADK executes only the last one; the earlier parallel calls are silently
dropped, produce no tool response, and leave the session in a mismatched state.

## 🔴 Required Information

**Describe the Bug:**
On the live/BIDI path (`Runner.runLive` + `LiveRequestQueue`), when the Gemini
Live API emits a single `BidiGenerateContentToolCall` message containing **multiple**
function calls (parallel tool calling), only the **last** call is executed. In
`GeminiLlmConnection.createToolCallResponse`, the response `Content` is rebuilt
**inside** the loop over the calls, so each iteration overwrites the previous one
and only the final `FunctionCall` survives:

```java
// GeminiLlmConnection.createToolCallResponse (before)
toolCall
    .functionCalls()
    .ifPresent(
        calls -> {
          for (FunctionCall call : calls) {
            builder.content(                                   // overwrites every iteration
                Content.builder()
                    .parts(ImmutableList.of(Part.builder().functionCall(call).build()))
                    .build());                                 // and no role is set
          }
        });
```

Two problems follow:
1. All parallel calls except the last are discarded before the `Event` is even
   built, so downstream code (including `handleFunctionCallsLive`, which itself
   loops correctly) never sees them. No `functionResponse` is generated for the
   dropped calls.
2. The built `Content` has no `role`, so it can later be treated as empty/invalid
   in session history.

The trigger is **any** single message carrying ≥2 calls — including several calls to
the **same** tool (e.g. `getWeather` for two different cities), not only calls to
distinct tools. A single call per message is unaffected (one loop iteration has
nothing to overwrite).

Because the drop happens at the connection layer, it is invisible from the ADK
`Event` stream (the resulting `Event` only ever shows the surviving last call),
which makes it easy to mistake for the model simply calling one tool.

This is specific to the **Gemini live connection** — the non-live `runAsync` path
(`Gemini.generateContent` → `handleFunctionCalls`) is unaffected and handles
parallel calls correctly.

**Steps to Reproduce:**
1. Build an `LlmAgent` with two independent tools (e.g. `getWeather(city)` and
   `getTime(timezone)`) and a live-capable Gemini model.
2. Start a live session (`runner.runLive(userId, sessionId, liveRequestQueue, runConfig)`).
3. Send one user turn that requires both tools (e.g. "What's the weather in Paris
   and the time in Europe/London? Call both tools in the same turn.").
4. Enable `DEBUG` logging for `com.google.adk.models.GeminiLlmConnection` to see the
   raw server `toolCall` message, and record which tools actually executed.

**Expected Behavior:**
All function calls contained in a single `toolCall` message are preserved and
executed. The built content carries `role = "model"` and remains in session
history, so every call gets a matching `functionResponse` and the turn completes
normally.

**Observed Behavior:**
Only the last function call is kept. The raw server message contains 2 calls, but
the `Event` ADK builds from it contains only the last one — so only that tool runs;
the earlier call has no `functionResponse`. The model then waits for a tool result
that will never be produced, so the turn never completes and the live session hangs
(surfaced as a `TimeoutException` once the client stops waiting). The dropped call in
the `Event` is the direct evidence; the hang is a downstream symptom of the missing
response. Single-call turns are unaffected — with one call the loop's single
iteration has nothing to overwrite, which is why the bug stayed latent.

**Environment Details:**

 - ADK Library Version (see maven dependency): `1.6.1-SNAPSHOT` (reproduced on `1.6.0`)
 - OS: Windows 11 (transport-independent; reproducible on macOS/Linux)
 - TS Version (tsc --version): N/A (Java) — JDK 17

**Model Information:**

 - Which model is being used: `gemini-3.1-flash-live-preview` (any Gemini
   Live/`bidiGenerateContent` model that emits parallel calls; response modality
   `AUDIO`)

---

## 🟡 Optional Information

**Regression:**
No — the loop-overwrite has been present in `createToolCallResponse` since the live
connection was introduced.

**Logs:**
With `DEBUG` logging on `com.google.adk.models.GeminiLlmConnection`, the raw Live
server message carries **both** calls, but the `Event` ADK builds from it contains
only the **last** one (and no `role`):

```text
# Raw server toolCall message (DEBUG: "Received server message") - TWO calls
{"toolCall":{"functionCalls":[
  {"args":{"city":"Paris"},"name":"getWeather"},
  {"args":{"timezone":"Europe/London"},"name":"getTime"}]}}

# The Event ADK constructed - only ONE call survived, and no "role"
Event: {"content":{"parts":[{"functionCall":{"args":{"timezone":"Europe/London"},"name":"getTime"}}]}, ...}

# getWeather never gets a functionResponse. The model waits for it, so the turn
# never completes (no turnComplete) and the session hangs until the client gives up:
(runLive ended: java.util.concurrent.TimeoutException)
```

The direct evidence of the drop is the `Event` above (`getWeather` is gone before any
tool execution is attempted); the timeout is a downstream symptom of the missing
response.

**Additional Context:**
Multiple calls per `toolCall` message is by design, not an edge case:
`LiveServerToolCall.functionCalls()` is a `List<FunctionCall>`, the live endpoint does
emit several calls in one message (parallel function calling), and the downstream
`handleFunctionCallsLive` already fans out over all calls and merges the results
(`mergeParallelFunctionResponseEvents`). Only `createToolCallResponse` fails to keep
them all.

The defect can be exercised directly via the package-private
`GeminiLlmConnection.convertToServerResponse` with a `LiveServerMessage` whose
`toolCall` carries two `FunctionCall`s — before the fix, the resulting `LlmResponse`
content has only 1 part instead of 2:

```java
LiveServerToolCall toolCall =
    LiveServerToolCall.builder()
        .functionCalls(ImmutableList.of(
            FunctionCall.builder().name("getWeather").build(),
            FunctionCall.builder().name("getTime").build()))
        .build();
LiveServerMessage message = LiveServerMessage.builder().toolCall(toolCall).build();

List<LlmResponse> out =
    GeminiLlmConnection.convertToServerResponse(message).toList().blockingGet();
// before fix: out.get(0).content().get().parts().get().size() == 1  (getTime only)
// after fix : == 2  (getWeather + getTime), role == "model"
```

 - Always (100%) — deterministic whenever a single `toolCall` message carries ≥2
   function calls (i.e. whenever the model parallel-calls in a live turn).

svetanis:

- Closes: #1357 (https://github.com/google/adk-java/issues/1357)

**Problem:**
On the live/BIDI path, when the Gemini Live API returns multiple function calls in
a single `toolCall` message (parallel tool calling), only the **last** call is
executed. `GeminiLlmConnection.createToolCallResponse` rebuilds the response
`Content` **inside** the loop over the calls, so each iteration overwrites the
previous content and only the final `FunctionCall` survives. The earlier calls are
dropped before the `Event` is built, never receive a `functionResponse`, and the
constructed content also lacks a `role` (so it can later be discarded as empty).
The result is a mismatched tool exchange and a live turn that stalls waiting for a
response that is never produced. The non-live `runAsync` path is unaffected.

**Solution:**
Collect **all** function calls from the `toolCall` message into a single
`role = "model"` `Content` (one `Part` per call) instead of overwriting the content
on each iteration:

```java
// GeminiLlmConnection (after)
private static LlmResponse createToolCallResponse(LiveServerToolCall toolCall) {
  LlmResponse.Builder builder = LlmResponse.builder();
  toolCall
      .functionCalls()
      .filter(Predicate.not(List::isEmpty))
      .map(GeminiLlmConnection::toModelContent)
      .ifPresent(builder::content);
  return builder.partial(false).turnComplete(false).build();
}

private static Content toModelContent(List<FunctionCall> functionCalls) {
  ImmutableList<Part> parts =
      functionCalls.stream()
          .map(call -> Part.builder().functionCall(call).build())
          .collect(toImmutableList());
  return Content.builder().role("model").parts(parts).build();
}
```

This preserves every call in the message (different tools or the same tool repeated),
so `handleFunctionCallsLive` can execute all of them and emit a `functionResponse` for
each, and sets the required `role` — matching the non-live `runAsync` path. The
`filter` keeps the original behaviour for an empty `functionCalls` list (no content).
(Single-call turns were already fine — one loop iteration has nothing to overwrite —
which is why the bug stayed latent.)

### Testing Plan

**Unit Tests:**

Added tests to `GeminiLlmConnectionTest` covering the full range of what a single
`toolCall` message can carry:

- `convertToServerResponse_withMultipleParallelToolCalls_keepsAllFunctionCalls` —
  two calls to **different** tools; both survive, in order, with `role == "model"`.
- `convertToServerResponse_withParallelCallsToSameTool_keepsAllFunctionCalls` — two
  calls to the **same** tool with different args; both survive.
- `convertToServerResponse_withEmptyToolCall_emitsResponseWithoutContent` — an empty
  `functionCalls` list produces no content (preserving the original behaviour).

The existing single-call test
(`convertToServerResponse_withToolCall_mapsContentWithFunctionCall`) continues to
pass (now also asserting `role == "model"`). Before the fix, the two multi-call tests
fail (`parts` has size 1, not 2).

```text
JUnit version 4.13.2
..............
Time: 0.416
OK (14 tests)
```

Verified with a live agent that has two independent tools (`getWeather`, `getTime`),
driven via `runLive` and prompted to call both in one turn (Gemini Developer API key,
no Vertex; a `bidiGenerateContent` model such as `gemini-3.1-flash-live-preview`).
With `DEBUG` logging on `GeminiLlmConnection`, the raw Live server message carries
both calls:

```text
Received server message: {"toolCall":{"functionCalls":[
  {"args":{"city":"Paris"},"name":"getWeather"},
  {"args":{"timezone":"Europe/London"},"name":"getTime"}]}}
```

- **Before the fix:** the `Event` ADK builds from that message contains only the last
  call (`getTime`); `getWeather` receives no `functionResponse`, so the turn never
  completes — the model waits for the missing response and the session hangs
  (`java.util.concurrent.TimeoutException`).
- **After the fix:** both calls appear in the model-role `Event`, both tools run,
  both `functionResponse`s are produced, and the turn completes normally.

### Checklist

### Additional context
The drop happens at the connection layer, before any ADK `Event` is built, so it is
invisible from the `Event` stream alone — the surviving last call looks like a normal
single tool call, and the resulting hang surfaces only as a `TimeoutException`.
Diagnosing it therefore means inspecting the raw Live server `toolCall` message
(logged at `DEBUG`), which shows both calls. Multiple calls per `toolCall` message is
by design: `LiveServerToolCall.functionCalls()` is a `List`, and
`handleFunctionCallsLive` already fans out over all calls and merges the results —
only `createToolCallResponse` failed to keep them.

### DIFF
--- a/core/src/main/java/com/google/adk/models/GeminiLlmConnection.java
+++ b/core/src/main/java/com/google/adk/models/GeminiLlmConnection.java
@@ -47,6 +47,7 @@ import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Predicate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -201,16 +202,24 @@ public final class GeminiLlmConnection implements BaseLlmConnection {
     LlmResponse.Builder builder = LlmResponse.builder();
     toolCall
         .functionCalls()
-        .ifPresent(
-            calls -> {
-              for (FunctionCall call : calls) {
-                builder.content(
-                    Content.builder()
-                        .parts(ImmutableList.of(Part.builder().functionCall(call).build()))
-                        .build());
-              }
-            });
+        .filter(Predicate.not(List::isEmpty))
+        .map(GeminiLlmConnection::toModelContent)
+        .ifPresent(builder::content);
     return builder.partial(false).turnComplete(false).build();
+  }
+
+  /**
+   * Wraps all function calls from a single tool-call message into one {@code "model"} content.
+   *
+   * <p>Every call must be kept (a message may carry multiple parallel calls) and the role is
+   * required so the content is not later discarded as empty.
+   */
+  private static Content toModelContent(List<FunctionCall> functionCalls) {
+    ImmutableList<Part> parts =
+        functionCalls.stream()
+            .map(call -> Part.builder().functionCall(call).build())
+            .collect(toImmutableList());
+    return Content.builder().role("model").parts(parts).build();
   }
 
   private static LlmResponse createUsageMetadataResponse(UsageMetadata usageMetadata) {
--- a/core/src/test/java/com/google/adk/models/GeminiLlmConnectionTest.java
+++ b/core/src/test/java/com/google/adk/models/GeminiLlmConnectionTest.java
@@ -19,6 +19,7 @@ package com.google.adk.models;
 import static com.google.common.truth.Truth.assertThat;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.genai.types.Content;
 import com.google.genai.types.FunctionCall;
 import com.google.genai.types.GenerateContentResponseUsageMetadata;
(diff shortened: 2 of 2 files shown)


hemasekhar-p:

kvmilos:
Thanks @svetanis  — this got fixed on `main` via an internal change that mirrors your fix (all parallel calls preserved, role="model"). Your report and PR nailed it. I'll close this as fixed; wanted to credit you directly first. Thanks for the careful write-up.

svetanis:
Thanks, @kvmilos! 

I really appreciate you crediting me directly and confirming the diagnosis
before closing. Glad the fix landed on main: preserving all parallel function
calls with `role="model"` is exactly the behavior that matters. 

No concerns about routing it through your internal change — 
the outcome is what counts, and it's a good one. 

Thanks for the careful review; I'll keep an eye out for anything similar. 
Cheers!

PR #1361, opened 2026-07-19
fix(firestore): honor appName in getSession/deleteSession/listEvents

petrmarinec:
## Description of Change

`FirestoreSessionService` stores each session at `/{root}/{userId}/sessions/{sessionId}` — keyed only by `(userId, sessionId)`. `appName` is saved as a field (`APP_NAME_KEY`) but is not part of the key, and three methods fetch the document by `(userId, sessionId)` without checking it:

- `getSession(appName, userId, sessionId, ...)` — returns the document regardless of `appName`.
- `listEvents(appName, userId, sessionId)` — checks only that the document exists.
- `deleteSession(appName, userId, sessionId)` — deletes regardless of `appName`.

`listSessions(appName, userId)` already scopes correctly with `getSessionsCollection(userId).whereEqualTo(APP_NAME_KEY, appName)`, so the omission in the other three is an oversight.

Because `appName` is accepted but ignored, when several ADK applications share one Firestore for the same users, application B can read, list the events of, or delete application A's session by using A's session ID with any `appName` — e.g. `getSession("appB", U, S)` returns app A's session. Session `state` can hold OAuth tokens (via `SessionStateCredentialService`) and conversation history, so this is a cross-application read (confidentiality) and delete (integrity) issue. `InMemorySessionService` keys on `appName -> userId -> sessionId`, so this also brings the Firestore backend in line with the in-memory contract.

This change adds the `appName` check to `getSession`, `deleteSession`, and `listEvents`, treating a mismatch as "session not found", mirroring `listSessions`.

## Testing Plan

Scope check only; matching `appName` behaves exactly as before. Suggested regression test (Firestore emulator): create a session under `(appA, U, S)`, then assert `getSession(appB, U, S)` and `listEvents(appB, U, S)` raise `SessionNotFoundException` and `deleteSession(appB, U, S)` leaves it intact, while all three succeed for `appA`.

### DIFF
--- a/contrib/firestore-session-service/src/main/java/com/google/adk/sessions/FirestoreSessionService.java
+++ b/contrib/firestore-session-service/src/main/java/com/google/adk/sessions/FirestoreSessionService.java
@@ -177,6 +177,16 @@ public class FirestoreSessionService implements BaseSessionService {
                 return Maybe.empty();
               }
 
+              // Enforce the (appName, userId, sessionId) scope. The Firestore
+              // document is keyed only by (userId, sessionId), so without this
+              // check a caller could read a session that belongs to a different
+              // application for the same user. Treat an appName mismatch as
+              // "not found" so cross-application existence is not leaked.
+              if (!appName.equals(data.get(APP_NAME_KEY))) {
+                logger.warn("Session {} does not belong to app {}", sessionId, appName);
+                return Maybe.error(new SessionNotFoundException("Session not found: " + sessionId));
+              }
+
               // Fetch events based on config
               GetSessionConfig config =
                   configOpt.orElseGet(() -> GetSessionConfig.builder().build());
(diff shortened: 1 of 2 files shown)


hemasekhar-p:
Hi @petrmarinec, thank you for your contribution! We appreciate you taking the time to submit this pull request. To proceed with the review, could you please address the following,

2. As per our contribution policy, please squash your changes into a single commit.
3. Please resolve the current Maven build failures.

petrmarinec:
Thanks for the review! I've addressed all three points:

- **Unit tests:** added `getSession` / `deleteSession` / `listEvents` tests for the appName-mismatch case (a mismatched appName is treated as "not found" and does not delete), and updated the existing tests that now need to stub the stored `appName`.
- **Squashed** into a single commit.
- **Build failures resolved** — they came from the existing tests not stubbing `appName`; `mvn -pl contrib/firestore-session-service -am test` now passes (47 tests in `FirestoreSessionServiceTest`) and `fmt:check` is clean.

hemasekhar-p:
@petrmarinec, thank you for addressing those points and providing the updated tests. Your changes are now under review by our team and we will keep you posted if we require any further updates. Thanks you.

kvmilos (approved):

PR #1378, opened 2026-07-26
fix(runner): build a new message when saving input blobs, instead of writing into the caller's Content

### LINKED ISSUE #1377: [BUG] saveInputBlobsAsArtifacts throws on an immutable Content parts list

## 🔴 Required Information

**Describe the Bug:**

With `saveInputBlobsAsArtifacts(true)`, `Runner` replaces each inline blob with a placeholder by writing
into the parts list of the `Content` the caller passed to `runAsync`:

```java
// Runner.java:376-382
newMessage.parts().get().set(i, Part.fromText("Uploaded file: " + fileName + " ..."));
```

That list is never copied on the way in, so whether the write succeeds depends on how the message was
built:

| Construction | Backing list | Result |
|---|---|---|
| `builder().parts(List.of(text, blob))` | caller's `List.of`, uncopied | **throws** |
| `builder().parts(textBuilder, blobBuilder)` | genai's own `ImmutableList` | **throws** |
| `Content.fromParts(text, blob)` | `Arrays.asList` | works |

**Steps to Reproduce:**

> **The construction is the trigger.** `Content.fromParts(...)` and `parts(Part...)` wrap
> `Arrays.asList`, whose slots are settable, so they will **not** reproduce this. (That is also why
> `RunnerTest.runner_executesSaveArtifactFlow` passes today — its helper uses the `Part...` overload.)

1. Build a message with an inline blob and an **immutable** parts list:
   `Content.builder().role("user").parts(List.of(Part.fromText("hello"), Part.fromBytes(bytes, "text/plain"))).build()`
2. Run it through any `Runner` (e.g. `InMemoryRunner`) with
   `RunConfig.builder().saveInputBlobsAsArtifacts(true).build()`.
3. It throws before the model is called.

The linked PR adds `RunnerTest.saveInputBlobsAsArtifacts_immutablePartsList_savesArtifactAndCompletes`,
which **fails on current `main`** and passes with the fix.

**Expected Behavior:**

The blob is saved as an artifact and the appended message carries the `"Uploaded file: ..."` placeholder —
the documented behaviour — however the caller built the `Content`.

**Observed Behavior:**

`UnsupportedOperationException` out of `Runner.appendNewMessageToSession`, upstream of the model call: no
model request, and no user event appended to the session. Captured on `main`:

```text
java.lang.UnsupportedOperationException
	at com.google.common.collect.ImmutableList.set(ImmutableList.java:544)
	at com.google.adk.runner.Runner.appendNewMessageToSession(Runner.java:379)
	at com.google.adk.runner.Runner.lambda$8(Runner.java:547)
	... (RxJava frames elided)
```

With `List.of(...)` the top frame is `ImmutableCollections$AbstractImmutableList.set` instead. 

**Environment Details:**

- ADK Library Version: `1.7.1-SNAPSHOT`
- OS: Windows 11 (not OS-specific)
- TS Version: N/A — Java

**Model Information:**

- Any — the failure precedes the model call. Reproduced with a stub and with `gemini-2.5-flash`.

---

## 🟡 Optional Information

```java
Part text = Part.fromText("hello");
Part blob = Part.fromBytes("payload".getBytes(UTF_8), "text/plain");
RunConfig config = RunConfig.builder().saveInputBlobsAsArtifacts(true).build();

// throws
runner.runAsync(userId, sessionId,
    Content.builder().role("user").parts(List.of(text, blob)).build(), config).blockingSubscribe();

// throws — no workaround; genai builds the ImmutableList itself
runner.runAsync(userId, sessionId,
    Content.builder().role("user").parts(text.toBuilder(), blob.toBuilder()).build(), config)
    .blockingSubscribe();

// works — Arrays.asList, settable slots
runner.runAsync(userId, sessionId, Content.fromParts(text, blob), config).blockingSubscribe();
```

**How often has this issue occurred?:** Always (100%) — deterministic for any parts list that rejects
`set`.

svetanis:

- Closes: [#1377](https://github.com/google/adk-java/issues/1377)

**Problem:**

With `saveInputBlobsAsArtifacts(true)`, `Runner.appendNewMessageToSession` replaces each inline blob with
a placeholder by writing into the parts list of the `Content` the caller passed to `runAsync`
([`Runner.java:376-382`](https://github.com/google/adk-java/blob/main/core/src/main/java/com/google/adk/runner/Runner.java#L376)).
That list is never copied on the way in, so for an immutable one the `set` throws
`UnsupportedOperationException` — before the model is called, and before anything is appended:

| Construction | Backing list | Result |
|---|---|---|
| `builder().parts(List.of(text, blob))` | caller's `List.of`, uncopied | **throws** |
| `builder().parts(textBuilder, blobBuilder)` | genai's own `ImmutableList` | **throws** |
| `Content.fromParts(text, blob)` | `Arrays.asList` | works |

**Solution:**

Rewrite a **copy** of the parts list and build a new `Content` for the event. The loop moves into a
helper returning both the prepared message and the saves it scheduled:

```java
private record OffloadedMessage(Content message, Completable artifactSaves) { ... }

private OffloadedMessage offloadInputBlobs(Session session, Content message, String invocationId) {
  // The runner directly saves the artifacts (if applicable) in the user message and replaces
  // the artifact data with a file name placeholder.
  List<Part> parts = new ArrayList<>(message.parts().get());
  ...
  return new OffloadedMessage(
      message.toBuilder().parts(ImmutableList.copyOf(parts)).build(), saveArtifactsFlow);
}
```

so `appendNewMessageToSession` becomes one decision:

```java
OffloadedMessage offloaded =
    this.artifactService != null && saveInputBlobsAsArtifacts
        ? offloadInputBlobs(session, newMessage, invocationContext.invocationId())
        : OffloadedMessage.unchanged(newMessage);
```

The functional change is one line — `new ArrayList<>(...)`, which always accepts `set`. Documented
behaviour is unchanged: the persisted message still carries the placeholder, the blob is still saved, and
the model still never sees it. 

### Testing Plan

**Unit Tests:**

```text
Tests run: 94, Failures: 0, Errors: 0, Skipped: 0     # with the fix
Tests run: 94, Failures: 5                            # reverting only Runner.java
```

Nine tests added. Five fail without the fix (regression tests); four pass in both states (parity tests
pinning behaviour that must not change):

| Test | Without fix |
|---|---|
| `..._immutablePartsList_savesArtifactAndCompletes` | **fails** |
| `..._partBuilderPartsList_savesArtifactAndCompletes` | **fails** |
| `..._appendedEventReplacesBlobWithPlaceholder` | **fails** |
| `..._doesNotModifyCallerMessage` | **fails** |
| `..._storesBlobVerbatim` | **fails** |
| `..._fromPartsConstruction_savesArtifactAndCompletes` | passes |
| `..._disabled_keepsBlobAndSavesNothing` | passes |
| `..._textOnlyMessage_passesThroughUnchanged` | passes |
| `..._disabledWithTextOnlyMessage_passesThroughUnchanged` | passes |

The last two cover ordinary traffic: a text-only prompt with the option on (the runner still walks the
parts and, after the fix, copies them), and the default path with the option off (the rewrite is skipped
entirely and the caller's message passes through uncopied).

A demo drives one real agent turn per row through `InMemoryRunner` (no custom wiring) against
`gemini-2.5-flash`, varying only the construction and the flag, and reads back both the artifact store
and the session afterwards.

_Before:_

```text
Run A (immutable List.of)      threw UnsupportedOperationException   model calls 0   nothing appended
Run B (Part.Builder varargs)   threw UnsupportedOperationException   model calls 0   nothing appended
Run C (Content.fromParts)      completed, blob offloaded, placeholder appended
Run D (immutable, flag off)    completed, blob reaches the model, no artifact
BUG REPRODUCED
```

_After (same demo):_

```text
Run A (immutable List.of)      completed   model calls 1   payload back verbatim   placeholder appended
Run B (Part.Builder varargs)   completed   model calls 1   payload back verbatim   placeholder appended
Run C (Content.fromParts)      unchanged
Run D (immutable, flag off)    unchanged
Run E (text only, flag on)     completed, nothing stored, text through untouched
Run F (text only, flag off)    completed, nothing stored, text through untouched
FIXED
```

On every row where the offload ran, the payload came back out of storage verbatim, the model was never
shown the blob, and the model **was** shown the placeholder — so this is not merely "stopped crashing".

### Checklist

### DIFF
--- a/core/src/main/java/com/google/adk/runner/Runner.java
+++ b/core/src/main/java/com/google/adk/runner/Runner.java
@@ -347,6 +347,9 @@ public class Runner {
   /**
    * Appends a new user message to the session history with optional state delta.
    *
+   * <p>{@code newMessage} is never modified; when inline blobs are saved as artifacts, the appended
+   * event carries a copy in which the blob data is replaced by placeholders.
+   *
    * @throws IllegalArgumentException if message has no parts.
    */
   private Single<Event> appendNewMessageToSession(
@@ -357,12 +360,16 @@ public class Runner {
       @Nullable Map<String, Object> stateDelta) {
     checkArgument(newMessage.parts().isPresent(), "No parts in the new_message.");
 
+    Content messageToAppend = newMessage;
     Completable saveArtifactsFlow = Completable.complete();
     if (this.artifactService != null && saveInputBlobsAsArtifacts) {
       // The runner directly saves the artifacts (if applicable) in the user message and replaces
-      // the artifact data with a file name placeholder.
-      for (int i = 0; i < newMessage.parts().get().size(); i++) {
-        Part part = newMessage.parts().get().get(i);
+      // the artifact data with a file name placeholder. The rewrite happens on a copy of the parts
+      // list: the caller's list may be immutable, and the caller does not expect the message it
+      // passed to runAsync to be modified.
+      List<Part> parts = new ArrayList<>(newMessage.parts().get());
+      for (int i = 0; i < parts.size(); i++) {
+        Part part = parts.get(i);
         if (part.inlineData().isEmpty()) {
           continue;
         }
@@ -373,14 +380,11 @@ public class Runner {
                     .saveArtifact(this.appName, session.userId(), session.id(), fileName, part)
                     .ignoreElement());
 
-        newMessage
-            .parts()
-            .get()
-            .set(
-                i,
-                Part.fromText(
-                    "Uploaded file: " + fileName + ". It has been saved to the artifacts"));
+        parts.set(
+            i,
+            Part.fromText("Uploaded file: " + fileName + ". It has been saved to the artifacts"));
       }
+      messageToAppend = newMessage.toBuilder().parts(ImmutableList.copyOf(parts)).build();
     }
     // Appends only. We do not yield the event because it's not from the model.
     Event.Builder eventBuilder =
@@ -388,7 +392,7 @@ public class Runner {
             .id(Event.generateEventId())
             .invocationId(invocationContext.invocationId())
             .author("user")
-            .content(newMessage);
+            .content(messageToAppend);
 
     // Add state delta if provided
     if (stateDelta != null && !stateDelta.isEmpty()) {
--- a/core/src/test/java/com/google/adk/runner/RunnerTest.java
+++ b/core/src/test/java/com/google/adk/runner/RunnerTest.java
@@ -23,6 +23,7 @@ import static com.google.adk.testing.TestUtils.createTestAgentBuilder;
 import static com.google.adk.testing.TestUtils.createTestLlm;
 import static com.google.adk.testing.TestUtils.createTextLlmResponse;
 import static com.google.adk.testing.TestUtils.simplifyEvents;
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.truth.Truth.assertThat;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Arrays.stream;
@@ -50,6 +51,7 @@ import com.google.adk.agents.SequentialAgent;
 import com.google.adk.apps.App;
 import com.google.adk.apps.ResumabilityConfig;
 import com.google.adk.artifacts.BaseArtifactService;
+import com.google.adk.artifacts.InMemoryArtifactService;
 import com.google.adk.events.Event;
 import com.google.adk.flows.llmflows.Functions;
 import com.google.adk.models.LlmRequest;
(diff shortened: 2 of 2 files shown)


hemasekhar-p:

core/src/main/java/com/google/adk/runner/Runner.java
```diff
@@ -344,9 +344,61 @@ public Completable close() {
...
   }
 
+  /**
+   * A user message prepared for persistence, together with the artifact saves it scheduled.
+   *
+   * @param message the message to append: a copy when inline blobs were replaced by placeholders,
+   *     otherwise the original
+   * @param artifactSaves the saves to run before the event is appended
+   */
+  private record OffloadedMessage(Content message, Completable artifactSaves) {
```
MiloszSobczyk: Adding this class seems like a lot of machinery than the fix needs. The code provided below fixes the issue without introducing `record` classes. Please consider changing the current implementation,
```java 
/**
   * Appends a new user message to the session history with optional state delta.
   *
   * <p>{@code newMessage} is never modified; when inline blobs are saved as artifacts, the appended
   * event carries a copy in which the blob data is replaced by placeholders.
   *
   * @throws IllegalArgumentException if message has no parts.
   */
  private Single<Event> appendNewMessageToSession(
      Session session,
      Content newMessage,
      InvocationContext invocationContext,
      boolean saveInputBlobsAsArtifacts,
      @Nullable Map<String, Object> stateDelta) {
    checkArgument(newMessage.parts().isPresent(), "No parts in the new_message.");

    Content messageToAppend = newMessage;
    Completable saveArtifactsFlow = Completable.complete();
    if (this.artifactService != null && saveInputBlobsAsArtifacts) {
      // The runner directly saves the artifacts (if applicable) in the user message and replaces
      // the artifact data with a file name placeholder. The rewrite happens on a copy of the parts
      // list: the caller's list may be immutable, and the caller does not expect the message it
      // passed to runAsync to be modified.
      List<Part> parts = new ArrayList<>(newMessage.parts().get());
      for (int i = 0; i < parts.size(); i++) {
        Part part = parts.get(i);
        if (part.inlineData().isEmpty()) {
          continue;
        }
        String fileName = "artifact_" + invocationContext.invocationId() + "_" + i;
        saveArtifactsFlow =
            saveArtifactsFlow.andThen(
                this.artifactService
                    .saveArtifact(this.appName, session.userId(), session.id(), fileName, part)
                    .ignoreElement());

        parts.set(
            i,
            Part.fromText("Uploaded file: " + fileName + ". It has been saved to the artifacts"));
      }
      messageToAppend = newMessage.toBuilder().parts(ImmutableList.copyOf(parts)).build();
    }
    // Appends only. We do not yield the event because it's not from the model.
    Event.Builder eventBuilder =
        Event.builder()
            .id(Event.generateEventId())
            .invocationId(invocationContext.invocationId())
            .author("user")
            .content(messageToAppend);

    // Add state delta if provided
    if (stateDelta != null && !stateDelta.isEmpty()) {
      eventBuilder.actions(
          EventActions.builder().stateDelta(new ConcurrentHashMap<>(stateDelta)).build());
    }

    return saveArtifactsFlow.andThen(
        this.sessionService.appendEvent(session, eventBuilder.build()));
  }```

File: core/src/test/java/com/google/adk/runner/RunnerTest.java [hunk 2]
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Google LLC
+ * Copyright 2026 Google LLC
```
MiloszSobczyk: Year in copyright notice indicates the time file was created, not last modified. Please revert to 2025.

core/src/test/java/com/google/adk/runner/RunnerTest.java
MiloszSobczyk, line 1: Every new test reuses the same 3 lines of code at the beginning. Please consider introducing the helper function to reduce code repetition.

MiloszSobczyk (changes requested):
Hi @svetanis, thank you for your contribution. Please look into the comments I left within my review.

svetanis:
Thanks for the review @MiloszSobczyk. 
All three comments addressed:
1. Removed the `OffloadedMessage` record and `offloadInputBlobs`; the offload is inline in
   `appendNewMessageToSession`, as in your comment. The fix itself is unchanged:
   `new ArrayList<>(newMessage.parts().get())`.
2. Copyright year reverted to 2025.
3. The repeated setup in the new tests is now a single `useRunnerWithArtifactService()`
   helper, which points the existing `runner` and `session` fields at a runner backed by a
   fresh `InMemoryArtifactService`. 

The branch has also been rebased against the latest `main`.

MiloszSobczyk:
Thank you for addressing my comments @svetanis. I have two more suggestions regarding the test suite. Please consider following them.

1. `saveInputBlobsAsArtifacts_doesNotModifyCallerMessage` never asserts that the artifact was saved, so it would still pass if the feature silently became a no-op. Please add an `assertThat(artifactNames()).hasSize(1);`.

2. `saveInputBlobsAsArtifacts_appendedEventReplacesBlobWithPlaceholder` uses `var unused = ... .test()` with no `assertComplete()`. That contradicts the reasoning you wrote yourself in `storesBlobVerbatim`. Please assert completion here too.

svetanis:
Thanks for the review, @MiloszSobczyk. Both applied.

Changes in the `saveInputBlobsAsArtifacts_*` tests:
- `_doesNotModifyCallerMessage` — asserts the artifact count.
- `_appendedEventReplacesBlobWithPlaceholder` — asserts completion, the role and the text part;
  placeholder matched exactly against the saved artifact name.
- `_storesBlobVerbatim` — asserts the artifact count and the mime type.
- `_fromPartsConstruction_savesArtifactAndCompletes` — placeholder matched exactly against the
  saved artifact name.
- `_twoBlobs_namesEachArtifactAfterItsPartIndex` — new test.
- All ten — completion now goes through one `assertAgentReplied(...)` helper, which also asserts
  the agent's reply event.

MiloszSobczyk (approved):

sherryfox:
It is already merged with [80c1a21](https://github.com/google/adk-java/commit/80c1a21da378f121fef3af065eb65dce0e0080c9)

PR #1381, opened 2026-07-28
feat: expose turnComplete for live content

### LINKED ISSUE #1200: Expose turnComplete control for ADK Java live content sends

  Yes. ADK Java live sessions currently expose `LiveRequestQueue.content(Content)`, but there is no public way to control whether that content completes the current live turn.

  For normal user input, completing the turn is correct. But some live applications need to append content to an already-open live session without immediately asking the model to
  generate.

  The lower-level GenAI Java live API supports `turnComplete`, but ADK Java does not currently surface it through `LiveRequestQueue`, `LiveRequest`, or `BaseLlmConnection`.

  Please expose an additive API for live content sends that allows callers to set `turnComplete`.

  Example:

  ```java
  requestQueue.content(content, false);

  Existing behavior should remain unchanged:

  requestQueue.content(content); // defaults to current behavior

  ### Impact on your work

  We are building long-running Gemini Live / ADK Java sessions where backend services may need to update live session context without treating the update as a normal user transcript turn.

  Without this API, applications must work around the behavior by suppressing model output after certain content sends. That is less precise because the model may still be asked to
  generate even when the caller only wanted to append content.

  This is needed for current live-session work, but there is no hard public launch deadline.

  ### Willingness to contribute

  Yes. I am willing to submit a PR if the maintainers agree with the API direction.

  ———

  ## 🟡 Recommended Information

  1. Application-level suppression

  Track when special content is enqueued, suppress model output until the next turnComplete=true, and avoid transcript persistence for that output window.

  This works as a workaround, but ADK still asks the model to generate.

  2. Custom BaseLlmConnection

  Implement a custom Gemini live connection that sends LiveSendClientContentParameters.turnComplete(false).

  This would duplicate ADK internals and increase maintenance risk.

  3. Reflection or private API access

  Not recommended.

  One possible additive API:

  public final class LiveRequestQueue {
    public void content(Content content) {
      processor.onNext(LiveRequest.builder().content(content).build());
    }

    public void content(Content content, boolean turnComplete) {
      processor.onNext(
          LiveRequest.builder()
              .content(content)
              .turnComplete(turnComplete)
              .build());
    }
  }

  LiveRequest could carry:

  public abstract Optional<Boolean> turnComplete();

  The live send loop could dispatch:

  connection.sendContent(
      request.content().get(),
      request.turnComplete().orElse(true));

  BaseLlmConnection could preserve compatibility with a default method:

  default Completable sendContent(Content content, boolean turnComplete) {
    return sendContent(content);
  }

  Then GeminiLlmConnection can pass the flag through:

  LiveSendClientContentParameters.builder()
      .turns(ImmutableList.of(content))
      .turnComplete(turnComplete)
      .build();

  ### Additional Context

  This should be backward compatible if LiveRequestQueue.content(Content) continues to behave exactly as it does today, and only callers using the new overload opt into explicit turn-
  completion behavior.

  The lower-level live API already supports the concept; this request is mainly about exposing that capability through ADK Java's live queue abstraction.

YuqiGuo105:
## Summary

- expose `turnComplete` on queued Live content requests
- add `LiveRequestQueue.content(Content, boolean)` while preserving the existing default behavior
- propagate the flag through `BaseLlmFlow` and `BaseLlmConnection`
- forward the value to Gemini's `LiveSendClientContentParameters`
- cover the legacy default, explicit `false`, and JSON round-trip behavior

## Motivation

Gemini Live supports sending multiple content messages as one user turn. A caller can send intermediate content with `turnComplete=false`, then mark the final content with `turnComplete=true`.

ADK Java did not expose this capability. `LiveRequest` had no turn-completion field, `BaseLlmFlow` called only `sendContent(Content)`, and `GeminiLlmConnection` always set `turnComplete(true)`. As a result, applications could not append content to an open Live turn without asking the model to process a completed turn.

## Implementation

`LiveRequest` now carries an optional `turnComplete` value. `LiveRequestQueue` provides an additive overload:

```java
requestQueue.content(content, false);
```

The existing method remains equivalent to `content(content, true)`, so current callers keep the same behavior.

`BaseLlmConnection` adds a default `sendContent(Content, boolean)` overload that delegates to the original method. This allows existing custom connection implementations to continue compiling and running. `GeminiLlmConnection` overrides the overload and forwards the value to the lower-level Live API.

Pure `FunctionResponse` content continues to use the existing tool-response path; the new flag applies to normal client content.

## User impact

Applications can now send multiple content messages within one Live user turn:

```java
requestQueue.content(firstContent, false);
requestQueue.content(finalContent, true);
```

Existing `requestQueue.content(content)` calls remain unchanged.

## Validation

- `./mvnw -pl core -Dtest=LiveRequestQueueTest test`
- `./mvnw test`

Closes #1200.

### DIFF
--- a/core/src/main/java/com/google/adk/agents/LiveRequest.java
+++ b/core/src/main/java/com/google/adk/agents/LiveRequest.java
@@ -46,6 +46,16 @@ public abstract class LiveRequest extends JsonBaseModel {
   public abstract Optional<Content> content();
 
   /**
+   * Returns whether sending {@link #content()} should complete the current turn.
+   *
+   * <p>If unset, content sends complete the turn to preserve the existing behavior.
+   *
+   * @return An optional boolean indicating whether the current turn should be completed.
+   */
+  @JsonProperty("turnComplete")
+  public abstract Optional<Boolean> turnComplete();
+
+  /**
    * Returns the blob of the request.
    *
    * <p>If set, send the blob to the model in realtime mode.
@@ -82,6 +92,9 @@ public abstract class LiveRequest extends JsonBaseModel {
 
     @JsonProperty("content")
     public abstract Builder content(@Nullable Content content);
+
+    @JsonProperty("turnComplete")
+    public abstract Builder turnComplete(@Nullable Boolean turnComplete);
 
     @JsonProperty("blob")
     public abstract Builder blob(@Nullable Blob blob);
--- a/core/src/main/java/com/google/adk/agents/LiveRequestQueue.java
+++ b/core/src/main/java/com/google/adk/agents/LiveRequestQueue.java
@@ -38,7 +38,17 @@ public final class LiveRequestQueue {
   }
 
   public void content(Content content) {
-    processor.onNext(LiveRequest.builder().content(content).build());
+    content(content, true);
+  }
+
+  /**
+   * Sends content to the model with explicit control over whether it completes the current turn.
+   *
+   * @param content the content to send
+   * @param turnComplete whether the model should begin responding after receiving the content
+   */
+  public void content(Content content, boolean turnComplete) {
+    processor.onNext(LiveRequest.builder().content(content).turnComplete(turnComplete).build());
   }
 
   public void realtime(Blob blob) {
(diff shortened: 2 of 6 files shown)


hemasekhar-p:
Hi @YuqiGuo105, thank you for your contribution! We appreciate you taking the time to submit this pull request. Currently this PR is under review by our team, we will keep you posted if any additional information is required. thank you.

PR #1218, opened 2026-05-23
fix(agents): warn when AgentTool config_path escapes agent base directory

adilburaksen:
## Summary

`resolveSubAgentFromConfigPath` in `ConfigAgentUtils.java` resolves a sub-agent
`config_path` from an agent YAML. Absolute paths and relative paths containing `..` are
resolved as-is, so a `config_path` can point outside the agent's base directory without any
signal that this is happening.

Per maintainer guidance, blocking these paths outright is a breaking change (existing tools
legitimately rely on absolute / parent-relative `config_path` values), so this PR keeps the
current behavior and adds an observable warning ahead of stricter enforcement in a future
major release.

## Change

* Preserve backward-compatible resolution — absolute `config_path` values and relative
  paths containing `..` are still honored.
* Log a `WARN` deprecation notice when the resolved path escapes the agent's base
  directory:

  ```
  AgentTool config_path '{}' accesses a path outside the agent base directory; this
  behavior is deprecated and will be disallowed in a future release.
  ```

## Tests

`ConfigAgentUtilsTest` (50/50 passing):

* Absolute `config_path` still resolves successfully (backward compatibility).
* Relative `config_path` with `..` that escapes the base directory still resolves
  successfully and triggers the deprecation warning.
* In-bounds relative and in-bounds `..` paths continue to resolve.

## Related

Same code path exists in `adk-python` (google/adk-python#5826) and `adk-go`.

PR #1407, opened 2026-08-07
fix(configurable): prevent path traversal in AgentTool config_path resolution

prasanna8585:
## Summary

`ConfigAgentUtils.resolveSubAgentFromConfigPath()` accepted absolute `config_path` values unconditionally, and for relative values only **logged a warning** when the resolved path escaped the agent's own base directory -- it did not block the load:

```
if (Path.of(configPath).isAbsolute()) {
  subAgentConfigPath = Path.of(configPath);   // accepted unconditionally
} else {
  subAgentConfigPath = configDir.resolve(configPath);
}
Path resolvedConfigPath = subAgentConfigPath.normalize().toAbsolutePath();
Path baseDir = configDir.normalize().toAbsolutePath();
if (!resolvedConfigPath.startsWith(baseDir)) {
  logger.warn(...);   // <-- warning only, no return/throw
}
if (!Files.exists(subAgentConfigPath)) { ... }
return fromConfig(subAgentConfigPath.toString());   // proceeds regardless
```

This is the **same vulnerability class** already hard-fixed with a **breaking change** in both other language ports:
- `adk-go`: https://github.com/google/adk-go/commit/604dd63e647b12a2f0025f68606d692d7ae24d7a ("BREAKING: an absolute config_path is no longer accepted")
- `adk-python`: commit `171ae9e`

This port (issue #1218) had instead chosen a warn-only deprecation, leaving the traversal fully exploitable today: a `config_path` such as `../../another_tenant/secret_agent.yaml` or an absolute path is loaded and parsed as a full agent configuration, with only a log line noting the escape.

## Reachability

`config_path` is a field in a subagent reference within an agent's own YAML config (`sub_agents: - config_path: ...`). In any deployment where different trust domains' agent configs are hosted under a shared root (e.g. a multi-tenant agent-hosting platform, or any scenario where config content can be influenced by a less-trusted party), this allows reading and loading arbitrary files reachable by the process as agent configuration -- outside the intended per-agent containment directory.

## Fix

- Reject absolute `config_path` values outright.
- Reject (rather than warn on) a resolved path that escapes the agent's base directory.
- Resolve symlinks on both sides where the paths exist (`toRealPath()`, falling back to the lexical path when the target doesn't yet exist, so a missing file is still reported as not-found rather than misclassified as a traversal) -- matching the symlink-escape protection in `adk-go`'s second commit for the same fix.

## Testing

- `fromConfig_subAgentConfigPathTraversal_throwsConfigurationException` -- the exact `../` escape case, confirmed rejected.
- `fromConfig_subAgentAbsoluteConfigPath_throwsConfigurationException` -- the absolute-path case, confirmed rejected.
- The existing `fromConfig_withSubAgents_createsHierarchy` test (a legitimate in-directory subagent reference) is unmodified and continues to pass, confirming no regression.

## Verification performed

Maven Central is not reachable in the environment I used to develop this fix, so I could not run `mvn test` locally. I instead dynamically confirmed both the vulnerability and the fix using a faithful, line-for-line transcription of the real method (standard JDK only, no external dependencies needed to exercise this specific logic):

- **Before the fix**: a crafted `../secret_dir/victim_secret.yaml` path logged the deprecation warning, then proceeded to read the target file's full content regardless.
- **After the fix**: the same input is rejected with `"Path traversal detected: ..."`; an absolute path is rejected with `"Absolute paths are not allowed..."`; a legitimate in-directory reference still succeeds.

### DIFF
--- a/core/src/main/java/com/google/adk/agents/ConfigAgentUtils.java
+++ b/core/src/main/java/com/google/adk/agents/ConfigAgentUtils.java
@@ -250,21 +250,37 @@ public final class ConfigAgentUtils {
     Path subAgentConfigPath;
 
     if (Path.of(configPath).isAbsolute()) {
-      subAgentConfigPath = Path.of(configPath);
-    } else {
-      subAgentConfigPath = configDir.resolve(configPath);
+      throw new ConfigurationException(
+          "Absolute paths are not allowed in AgentTool config_path: " + configPath);
     }
+    subAgentConfigPath = configDir.resolve(configPath);
 
-    // Warn when the resolved config path escapes the agent's base directory. For backward
-    // compatibility this is still allowed, but the behavior is deprecated and will be disallowed
-    // in a future release.
+    // Reject a resolved config path that escapes the agent's own base directory.
+    // Both sides are made absolute and symlinks are resolved where the paths
+    // exist, so a symlink inside the agent directory cannot be used to escape
+    // it either. Matches the hard rejection adk-python (171ae9e) and adk-go
+    // (604dd63) ship for the identical AgentTool config_path resolution.
     Path resolvedConfigPath = subAgentConfigPath.normalize().toAbsolutePath();
     Path baseDir = configDir.normalize().toAbsolutePath();
-    if (!resolvedConfigPath.startsWith(baseDir)) {
-      logger.warn(
-          "AgentTool config_path '{}' accesses a path outside the agent base directory; this"
-              + " behavior is deprecated and will be disallowed in a future release.",
-          configPath);
+    try {
+      resolvedConfigPath = resolvedConfigPath.toRealPath();
+    } catch (IOException e) {
+      // Falls through with the lexical (non-symlink-resolved) path so a
+      // missing file still reports as "not found" below rather than as a
+      // traversal rejection.
+    }
+    Path realBaseDir = baseDir;
+    try {
+      realBaseDir = baseDir.toRealPath();
+    } catch (IOException e) {
+      // Base directory always exists (it contains the config file currently
+      // being loaded); this is defensive only.
+    }
+    if (!resolvedConfigPath.startsWith(realBaseDir)) {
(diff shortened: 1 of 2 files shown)


hemasekhar-p:
Hi @prasanna8585, thank you for your contribution! We truly appreciate for this fix. we have noticed that Your branch is currently out of sync with the main branch, and we heve encountered some issues with the Maven build. Could you please rebase your branch and resolve these build concerns so that we can move forward with the review process?

prasanna8585:
Hi @hemasekhar-p, thank you for the review.
i have rebase the branch can you review it again.

hemasekhar-p:
@prasanna8585 Thank you for the update. I noticed some inconsistencies in the code formatting now, could you please address those so we can proceed with the review?

prasanna8585:
Hi @hemasekhar-p, thanks for catching that - pushed a formatting fix, should be consistent now. Let me know if anything else needs adjusting.

hemasekhar-p:
@prasanna8585, could you please squash your commits into a single commit to keep the pull request history clean before we move forward with the review?

prasanna8585:
@hemasekhar-p, is there any update require from my side to merge this PR?

hemasekhar-p:
@prasanna8585, this PR is currently under review by our team. We will keep you updated if any changes are required. Thank you.

sherryfox:
Thanks @prasanna8585, and sorry for the outcome here — this is a timing call rather than anything wrong with the change.

Hardening this is on the roadmap: the deprecation warning on main (added in #1218) says the behaviour "will be disallowed in a future release", and we have an internal bug tracking that work pinned to the next major. Until then we're holding off on breaking changes to `config_path` resolution, since existing tools still rely on absolute and `..` paths — that was the conclusion in #1218 and it hasn't changed yet.

So I'm closing this one. The fix will be revisited as we approach the major release. Appreciate you picking this up though.

PR #1412, opened 2026-08-08
feat(plugins): add SaveFilesAsArtifactsPlugin, porting adk-python's file-upload offload

### LINKED ISSUE #1411: [FEATURE] Port SaveFilesAsArtifactsPlugin from adk-python

adk-python deprecated `save_input_blobs_as_artifacts` in favour of `SaveFilesAsArtifactsPlugin`.
adk-java has the parameter but not the plugin, so `RunConfig.saveInputBlobsAsArtifacts(true)` is the
only way to keep uploaded bytes out of every later LLM request — and it loses two things the plugin
provides.

**1. The uploaded file name is discarded.** `Runner` names every artifact
`artifact_{invocationId}_{index}` and never reads `Blob.displayName`.

```java
// Runner, in the blob-offload loop
String fileName = "artifact_" + invocationContext.invocationId() + "_" + i;
```

An upload of `report.pdf` is stored as `artifact_<invocationId>_1`, so anything that later loads it
must work from an opaque id that appears nowhere but the placeholder text.

**2. Nothing is reported to the session.** The user event carries a state delta only, so
`EventActions.artifactDelta` stays empty and the session's artifact bookkeeping never records the
upload.

```java
// Runner.appendNewMessageToSession — no artifactDelta is set on this event
if (stateDelta != null && !stateDelta.isEmpty()) {
  eventBuilder.actions(EventActions.builder().stateDelta(new ConcurrentHashMap<>(stateDelta)).build());
}
```
Both behaviours live in the plugin.

`SaveFilesAsArtifactsPlugin` in `com.google.adk.plugins`, registered on a `Runner` like any other
plugin. For each `inlineData` part of the incoming user message it would:

- save the part to the configured `BaseArtifactService`, named from `Blob.displayName`, falling back
  to `artifact_{invocationId}_{index}` when the blob carries no name;
- replace it with `[Uploaded Artifact: "<name>"]` in the message that reaches the model and is
  appended to the session, matching adk-python's wording;
- report the saved versions through `EventActions.artifactDelta`;
- on a failed save, keep the original part and log, without failing the invocation — as in
  adk-python, the log is the only signal, so an unreachable artifact service degrades silently to no
  offload.

**`attach_file_reference` would not be ported:** it needs
`get_artifact_version(...).canonical_uri`, which has no equivalent on `BaseArtifactService`, so it
would require an SPI change across every implementation. Separate change.

### Impact on your work

Applications wanting adk-python's behaviour must reimplement it in application code, where it will
drift from upstream. Not blocking, and no timeline — this is a parity gap, not an outage.

### Willingness to contribute

Yes. A PR follows immediately after this issue: one new plugin class plus a small package-private helper, with tests. No existing file is modified, and the only new public surface is the plugin class itself.

---

**Teach the existing flag to read `Blob.displayName` and set `artifactDelta`.** Smaller diff, but it
adds behaviour to the parameter adk-python is steering users away from, and adk-java would still have
no plugin — widening the divergence rather than closing it.

**Implement the plugin in application code.** This works using public API only, and is how the
behaviour below was verified. It is per-application boilerplate for something a plugin surface exists
to ship once.

Registration needs no new API — the existing surfaces already accept it:

```java
Runner runner = Runner.builder()
    .agent(agent)
    .appName("my-app")
    .artifactService(new InMemoryArtifactService())
    .sessionService(new InMemorySessionService())
    .plugins(new SaveFilesAsArtifactsPlugin())   // also App.Builder.plugins(...) and
    .build();                                    // InMemoryRunner(agent, appName, plugins)
```

Two hooks, both already wired into the runtime:

```java
public class SaveFilesAsArtifactsPlugin extends BasePlugin {

  @Override
  public Maybe<Content> onUserMessageCallback(InvocationContext ctx, Content userMessage) {
    // save each inlineData part, swap it for [Uploaded Artifact: "<name>"],
    // stash {fileName: version} under a temp: state key, return the rebuilt Content
  }

  @Override
  public Maybe<Content> beforeAgentCallback(BaseAgent agent, CallbackContext callbackContext) {
    // drain the stash into callbackContext.eventActions().artifactDelta(), return Maybe.empty()
  }
}
```

The state hand-off exists because `onUserMessageCallback` runs before any `EventActions` exists, so
the saved versions cannot be reported from there. adk-python solves it the same way.

### Additional Context

Same upload, same runner, differing only in which mechanism performs the offload:

```text
--- Run A: RunConfig.saveInputBlobsAsArtifacts(true) ---
  user attached                     : blob with displayName="report.pdf"
  artifacts stored in the session   : artifact_e-eb41628a-9812-43a0-88d0-7b121706190b_1
  EventActions.artifactDelta        : (none)
  message appended to the session   :
                                      - text: "read the attachment"
                                      - text: "Uploaded file: artifact_e-eb41628a-…_1. It has been saved to the artifacts"

--- Run B: SaveFilesAsArtifactsPlugin (the proposed port) ---
  user attached                     : blob with displayName="report.pdf"
  artifacts stored in the session   : report.pdf
  EventActions.artifactDelta        : {report.pdf=0}
  message appended to the session   :
                                      - text: "read the attachment"
                                      - text: "[Uploaded Artifact: "report.pdf"]"
```

Both offload the payload losslessly and both show the model a placeholder instead of the bytes — the
gap is the file name and the bookkeeping, not the offload.

Observed on `1.7.2-SNAPSHOT`, Windows 11 (not OS-specific), with a stub model and with
`gemini-3.5-flash`; both mechanisms run before the model call.

svetanis:

- Closes: #1411 (https://github.com/google/adk-java/issues/1411)

**Problem:**

adk-python deprecated `save_input_blobs_as_artifacts` in favor of `SaveFilesAsArtifactsPlugin`.
adk-java has the parameter but not the plugin, so `RunConfig.saveInputBlobsAsArtifacts(true)` is the
only way to keep uploaded bytes out of every later LLM request — and it loses two things:

1. **The uploaded file name is discarded.** `Runner` names every artifact
   `artifact_{invocationId}_{index}` and never reads `Blob.displayName`, so an upload of `report.pdf`
   is stored under an opaque id.
2. **Nothing is reported to the session.** The user event carries a state delta only, so
   `EventActions.artifactDelta` stays empty and the session's artifact bookkeeping never records the
   upload.

**Solution:**

A new `SaveFilesAsArtifactsPlugin` in `com.google.adk.plugins`, registered like any other plugin. For
each `inlineData` part of the incoming user message it saves the part to the configured
`BaseArtifactService` — named from `Blob.displayName`, falling back to the existing
`artifact_{invocationId}_{index}` — replaces it with `[Uploaded Artifact: "<name>"]` in the message
that reaches the model and the session, and reports the saved versions through
`EventActions.artifactDelta`. A failed save keeps the original part and logs, without failing the
invocation. The failure is reported only through the logger, matching adk-python — so an artifact
service that is misconfigured or unreachable degrades silently to no offload, with each upload
continuing to reach the model inline.

**No existing file is modified, and the only new public surface is the plugin class itself** — the
hand-off helper is package-private. Registration uses the three surfaces that already exist
(`Runner.Builder.plugins`, `App.Builder.plugins`, `InMemoryRunner`).

Four new files, all in `com/google/adk/plugins/`:

| File | Concern |
|---|---|
| `core/src/main/java/…/SaveFilesAsArtifactsPlugin.java` | the four hooks: save, name, replace, clean up on both the completing and the failing run |
| `core/src/main/java/…/PendingArtifactDelta.java` | package-private `onUserMessageCallback` → `beforeAgentCallback` hand-off |
| `core/src/test/java/…/SaveFilesAsArtifactsPluginTest.java` | 29 tests |
| `core/src/test/java/…/PendingArtifactDeltaTest.java` | 12 tests |

`PendingArtifactDelta` is package-private and tested on its own — no new package, no new public type.

Five deliberate divergences from adk-python:

- **The pending-delta stash key is invocation-scoped** (`temp:{plugin}:pending_delta:{invocationId}`).
  Python keys on the plugin name alone, so a stash stranded by a failed invocation is drained by the
  *next* invocation in that session, reporting artifacts it never saved. The invocation id makes a
  stranded entry inert, and the plugin clears its own stash on both exits — `afterRunCallback` when
  the run completes, `onRunErrorCallback` when it fails — so nothing is left behind to strand.
- **The stash is cleared by overwriting with an empty map, never `State.remove`.** `remove` writes the
  `State.REMOVED` sentinel into the event's state delta, which does not survive a JSON round trip.
- **With no artifact service configured the hook returns `Maybe.empty()`**, where Python returns the
  unmodified message. Same outcome for this plugin, but `PluginManager` early-exits on the first
  non-empty return — so Python's form also suppresses every *later* plugin's
  `on_user_message_callback` whenever this one is disabled. An empty `Maybe` cannot.
- **adk-python's 20 MB inline-data cap is not ported.** Upstream rejects an over-cap upload *before*
  saving it, so the file is dropped rather than offloaded. The cap guards the inline-request limit,
  but the blob has already been replaced by a placeholder here, so it never reaches the model either
  way. This port saves whatever it is handed and lets the artifact service decide what it can hold.

**`attach_file_reference` is deliberately not ported.** It needs
`get_artifact_version(...).canonical_uri`, which has no equivalent on `BaseArtifactService` — adding
it means changing the artifact-service SPI and every implementation. That is a separate, larger,
separately reviewable change. This port is placeholder-only, which is behaviorally what
`RunConfig.saveInputBlobsAsArtifacts` already does.

### Testing Plan

**Unit Tests:**

41 tests across two classes, mirroring `GlobalInstructionPluginTest`. Both contexts are mocked;
`State`, `Session` and `InMemoryArtifactService` are real, so the `temp:` state hand-off between the
hooks is exercised rather than stubbed.

```text
Tests run: 29, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.559 s -- in com.google.adk.plugins.SaveFilesAsArtifactsPluginTest
Tests run: 12, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.700 s -- in com.google.adk.plugins.PendingArtifactDeltaTest

[INFO] Results:
[INFO] Tests run: 41, Failures: 0, Errors: 0, Skipped: 0
```

`SaveFilesAsArtifactsPluginTest` (29) covers the display-name vs generated-name choice, the
placeholder swap, part order, multiple and duplicate uploads, the artifact delta and its drain, the
cleanup on both the completing and the failing run, the failed-save and partial-failure paths, both constructors, and that the caller's `Content` comes back untouched, for all three ways genai lets you build one.

`PendingArtifactDeltaTest` (12) covers the hand-off's own rules: the `temp:` key convention,
invocation and plugin scoping, clearing by overwrite rather than `State.remove`, reading back state
that has lost its type information, and that a second drain writes nothing — a write would set the
state delta, and `BaseAgent` emits an event for any before-agent callback that leaves one.

- [x] Verified end-to-end, against the offline stub and against `gemini-3.5-flash`.

A six-row harness builds a `Runner` explicitly (an `InMemoryRunner` cannot substitute a failing
artifact service) and registers the plugin. Below, **flag** means
`RunConfig.saveInputBlobsAsArtifacts(true)` and **plugin** means `SaveFilesAsArtifactsPlugin`.

| Row | Mechanism | Upload | Result |
|---|---|---|---|
| A | flag | named blob | stored as `artifact_<invocationId>_1`, `artifactDelta` **(none)** |
| B | plugin | named blob | stored as **`report.pdf`**, `artifactDelta` **`{report.pdf=0}`** |
| C | plugin | unnamed blob | falls back to `artifact_<invocationId>_1` — parity with the flag |
| D | plugin | text only | passes straight through |
| E | plugin, failing store | named blob | original part kept, **invocation still completes** |
| F | plugin + flag | named blob | plugin wins; one artifact, no double save, no crash |

A and B are the change, side by side:

```text
--- Run A: RunConfig.saveInputBlobsAsArtifacts(true) ---
  artifacts stored in the session   : artifact_e-eb41628a-9812-43a0-88d0-7b121706190b_1
  EventActions.artifactDelta        : (none)
  message appended to the session   : "Uploaded file: artifact_e-eb41628a-…_1. It has been saved to the artifacts"

--- Run B: SaveFilesAsArtifactsPlugin ---
  artifacts stored in the session   : report.pdf
  EventActions.artifactDelta        : {report.pdf=0}
  message appended to the session   : "[Uploaded Artifact: "report.pdf"]"
```

Both offload the payload losslessly and both show the model a placeholder — the change is the file
name and the bookkeeping, not the offload.

Row F is the migration row: the plugin strips every `inlineData` part before the runner's flag loop
sees the message, so enabling both is safe and there is no double save.

### Checklist

### DIFF
--- a/core/src/main/java/com/google/adk/plugins/PendingArtifactDelta.java
+++ b/core/src/main/java/com/google/adk/plugins/PendingArtifactDelta.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.adk.plugins;
+
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
+
+import com.google.adk.agents.CallbackContext;
+import com.google.adk.agents.InvocationContext;
+import com.google.common.collect.ImmutableMap;
+import java.util.Map;
+
+/**
+ * Carries artifact versions from {@code onUserMessageCallback}, which has no {@link
+ * com.google.adk.events.EventActions} to write to, across to {@code beforeAgentCallback}, which
+ * does.
+ *
+ * <p>The hand-off rides on the session state under a {@link
+ * com.google.adk.sessions.State#TEMP_PREFIX} key, so {@code BaseSessionService.appendEvent} skips
+ * it when applying the state delta and the bookkeeping never reaches persisted session state. The
+ * key also carries the invocation id, so concurrent invocations sharing one session cannot read
+ * each other's pending versions.
+ *
+ * <p>Draining overwrites the entry with an empty map rather than removing it. {@code State.remove}
+ * would write the {@code State.REMOVED} sentinel into the event's state delta, which does not
+ * survive a JSON round trip.
+ */
+final class PendingArtifactDelta {
+
+  private static final String KEY = "temp:%s:pending_delta:%s";
+
+  private PendingArtifactDelta() {}
+
+  /** Stores the versions produced during this invocation. */
+  static void stash(
+      InvocationContext invocationContext, String pluginName, ImmutableMap<String, Integer> delta) {
+    invocationContext
+        .session()
+        .state()
+        .put(key(pluginName, invocationContext.invocationId()), delta);
+  }
+
+  /**
+   * Returns the stored versions and clears them, so only the first agent callback reports them.
+   *
+   * <p>The write is guarded on there being something to clear, and the guard is load-bearing: a
+   * write through {@link CallbackContext#state()} sets the state delta, and {@code BaseAgent} emits
+   * an event for any before-agent callback that left one. Clearing unconditionally would therefore
+   * emit an event carrying an empty artifact delta for every agent after the first.
+   */
+  static ImmutableMap<String, Integer> drain(CallbackContext callbackContext, String pluginName) {
+    String key = key(pluginName, callbackContext.invocationId());
+    ImmutableMap<String, Integer> pending = read(callbackContext.state().get(key));
+    if (!pending.isEmpty()) {
+      callbackContext.state().put(key, ImmutableMap.of());
+    }
+    return pending;
+  }
+
+  /**
+   * Discards versions left behind by an invocation that ended before any agent ran, which happens
+   * when a {@code beforeRunCallback} on another plugin halts the run. There is no {@link
+   * com.google.adk.events.EventActions} to report them on at that point, so they are dropped rather
+   * than returned.
+   *
+   * <p>Only writes when something is actually stashed: an unconditional write would add an entry
+   * for every invocation that uploaded nothing, which is more state noise than the leak it
+   * prevents.
+   */
+  static void clear(InvocationContext invocationContext, String pluginName) {
+    Map<String, Object> state = invocationContext.session().state();
+    String key = key(pluginName, invocationContext.invocationId());
+    if (!read(state.get(key)).isEmpty()) {
+      state.put(key, ImmutableMap.of());
+    }
+  }
+
+  private static String key(String pluginName, String invocationId) {
+    return KEY.formatted(pluginName, invocationId);
+  }
+
+  private static ImmutableMap<String, Integer> read(Object stashed) {
+    if (!(stashed instanceof Map<?, ?> entries)) {
+      return ImmutableMap.of();
+    }
+    return entries.entrySet().stream()
+        .filter(PendingArtifactDelta::isVersionEntry)
+        .collect(
+            toImmutableMap(entry -> (String) entry.getKey(), entry -> (Integer) entry.getValue()));
+  }
+
+  /** State values survive a JSON round trip untyped, so each entry is checked before it is kept. */
+  private static boolean isVersionEntry(Map.Entry<?, ?> entry) {
+    return entry.getKey() instanceof String && entry.getValue() instanceof Integer;
+  }
+}
--- a/core/src/main/java/com/google/adk/plugins/SaveFilesAsArtifactsPlugin.java
+++ b/core/src/main/java/com/google/adk/plugins/SaveFilesAsArtifactsPlugin.java
@@ -1,0 +1,227 @@
+/*
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


hemasekhar-p:

PR #1419, opened 2026-08-10
fix(artifacts): do not derive a save's version number from a failed listing

### LINKED ISSUE #1418: [BUG] GcsArtifactService.saveArtifactAndReturnBlob overwrites version 0 when listVersions fails

### Describe the Bug

`GcsArtifactService.listVersions` returns an empty list when the listing raises `StorageException`:

```java
// core/src/main/java/com/google/adk/artifacts/GcsArtifactService.java — listVersions
} catch (StorageException e) {
  return ImmutableList.of();
}
```

`saveArtifactAndReturnBlob` derives the next version number from that result, and the write that
follows carries no precondition:

```java
// saveArtifactAndReturnBlob
return listVersions(appName, userId, sessionId, filename)
    .map(versions -> versions.isEmpty() ? 0 : max(versions) + 1)
// ...
Blob blob = storageClient.create(blobInfo, dataToSave);
```

At the point where the version number is chosen, an empty list has two possible meanings: the
artifact has no versions yet, or the listing did not complete. A 503, a 429, a timeout, or IAM
propagation after a binding change produces the second, and the save proceeds as if it were the
first — it computes version 0 and writes over the object already stored at version 0. The call
returns success and version `0`. The earlier content is no longer retrievable, and nothing is
logged.

Both conditions are needed for the loss: a save whose listing fails, **and** at least one version of
that artifact already stored. A listing failure on the first save of a file changes nothing — 0 is
the correct version and the write lands on an unused name. The computed version is always 0, so
higher versions are never overwritten.

**Example.** An agent saves `report.txt` at 10:00 and gets version 0. At 10:05 it saves again and
the listing fails for two seconds. Both calls returned success and version 0; `loadArtifact(..., 0)`
now returns the 10:05 content, and the 10:00 content is gone.

**With versions 0 and 1 both stored**, the same failure writes the newest content to version 0.
`loadArtifact` with no version resolves to `max(versions)`, so it then returns version 1 — the older
content.

How the file currently handles `StorageException`:

| Method | Operation | On failure |
|---|---|---|
| `loadArtifact` | `objects.get` | returns `null` |
| `listArtifactKeys` (session prefix) | `objects.list` | throws `VerifyException` |
| `listArtifactKeys` (user prefix) | `objects.list` | throws `VerifyException` |
| `deleteArtifact` | delete | throws `VerifyException` |
| `listVersions` | `objects.list` | returns `ImmutableList.of()` |
| `saveArtifactAndReturnBlob` | create | throws `VerifyException` |

`listArtifactKeys` and `listVersions` issue the same
`storageClient.list(bucketName, BlobListOption.prefix(...))` call. `listVersions` is the one whose
result selects the blob name a save writes to.

### Steps to Reproduce

Reproduced against a real GCS bucket. A transient cannot be scheduled, so the run reaches the same
branch deterministically by withholding `storage.objects.list` from the identity performing the
second save. The catch does not inspect the cause, so a 403 arrives there the same way a 503, a 429
or a socket error does. This is how the run triggers the branch, not a claim about how deployments
are configured — `storage.objects.list` is part of `roles/storage.objectViewer`, and artifact
identities normally hold it.

1. A bucket with uniform bucket-level access, and a service account granted `storage.objects.create`,
   `storage.objects.get` and `storage.objects.delete` on it, but not `storage.objects.list`. (Delete
   is what permits an overwrite in GCS at all; `objectAdmin` and `objectUser` both include it.)
2. A `GcsArtifactService` over a fully-permissioned `Storage` client:
   `saveArtifact("report.txt", "PAYLOAD-A")`.
3. A second `GcsArtifactService` over a `Storage` client authenticated as the service account from
   step 1: `saveArtifact("report.txt", "PAYLOAD-B")`. A transient affects one save, not every save.
4. `loadArtifact("report.txt", 0)`, then list `app/user/session/report.txt/` with the privileged
   credential.

The same path is reachable in a unit test without GCP: stub `Storage.list(...)` to throw
`StorageException`, save twice, and both saves compute version 0 and target the same blob name.

### Expected Behavior

A save whose version listing did not complete does not write, and the failure reaches the caller —
as it already does for the same operation in `listArtifactKeys`. The stored artifact is unchanged.

### Observed Behavior

Both saves return success and version `0`. The bucket holds a single object containing the second
payload, and `loadArtifact(..., 0)` returns it. The control arm — same code, same payloads, same
bucket, an identity that can list — returns versions 0 and 1 and keeps both objects.

```text
--- ARM 1 - listVersions denied ---
save 1 "report.txt" (client whose listing works) : PAYLOAD-A -> reported version 0
  bucket: 1 object   .../f18-session/report.txt/0 = PAYLOAD-A
save 2 "report.txt" (client under test)          : PAYLOAD-B -> reported version 0
  bucket: 1 object   .../f18-session/report.txt/0 = PAYLOAD-B
loadArtifact(version 0)                          : PAYLOAD-B

--- ARM 2 - control, identity that CAN list ---
save 1 "report.txt"                              : PAYLOAD-A -> reported version 0
save 2 "report.txt"                              : PAYLOAD-B -> reported version 1
  bucket: 2 objects  .../report.txt/0 = PAYLOAD-A, .../report.txt/1 = PAYLOAD-B
loadArtifact(version 0)                          : PAYLOAD-A
```

The arms differ only in which client performs the second save.

### Environment Details

- ADK Library Version: `1.7.2-SNAPSHOT`
- OS: not OS-specific

### Model Information

N/A — no model is involved.

---

### Related, reading the same result, not covered here

Two other callers derive behavior from the same empty list. Both misreport rather than overwrite,
and changing them is observable to callers, so they are noted here rather than folded into the fix:

| Site | Effect of the empty list |
|---|---|
| `loadArtifact`, no version given | `Maybe.empty()` — a stored artifact is reported as absent |
| `deleteArtifact` | `versions.isEmpty()` → `Completable.complete()` — a delete that removed nothing reports success |
| `dev` `ArtifactController.listArtifactVersions` | the endpoint returns `[]` and logs *"Found 0 versions for artifact X"* |

### Regression

No. The first revision of this file already returns an empty list on a failed listing and already
derives `nextVersion` from it.

### Willingness to contribute

Yes — a PR follows immediately, scoped to the save path. It gives that path its own view of the
listing: a private method that lets `StorageException` propagate, surfaced as
`VerifyException("Failed to list artifact versions from GCS", e)`, matching the wording
`listArtifactKeys` already uses. `listVersions` and its existing test are unchanged, so the three
callers above behave exactly as they do today.

### How often has this issue occurred?

Always (100%), when a save's version listing fails and at least one version is already stored.

svetanis:

- Closes: [#1418](https://github.com/google/adk-java/issues/1418)

**Problem:**

`GcsArtifactService.listVersions` returns an empty list when the listing raises `StorageException`.
`saveArtifactAndReturnBlob` derives the next version number from that result
(`versions.isEmpty() ? 0 : max(versions) + 1`) and writes with no precondition. At the point where
the version number is chosen, an empty list can mean either that the artifact has no versions or
that the listing did not complete — so a 503, a 429, a timeout or IAM propagation makes the save
compute version 0 and write over the object already stored at version 0, while returning success and
version `0` to the caller.

`listArtifactKeys` issues the same `storageClient.list(bucketName, BlobListOption.prefix(...))` call
and surfaces a failure as `VerifyException`. `listVersions` is the one whose result selects the blob
name a save writes to.

**Solution:**

Separate the listing from the decision about what a failed listing means. The query moves into a
private `readVersions` that lets `StorageException` propagate, and each caller applies its own
policy: `listVersions` catches it and returns an empty list as before, the save path catches it and
throws. Four pieces:

| # | Piece | What it is |
|---|---|---|
| 1 | `readVersions` — new private method | the old body of `listVersions`, moved verbatim; the stream pipeline is unchanged |
| 2 | `listVersions` — public, behavior unchanged | call `readVersions`, catch `StorageException`, return `ImmutableList.of()` |
| 3 | `versionsBeforeSaving` — new private method | call `readVersions`, catch `StorageException`, throw `VerifyException("Failed to list artifact versions from GCS", e)` — the same form `listArtifactKeys` already uses for this operation, naming what failed |
| 4 | `saveArtifactAndReturnBlob` — one line | was `listVersions(...)`, now `Single.fromCallable(() -> versionsBeforeSaving(...))`; everything after it is untouched |

```java
private ImmutableList<Integer> versionsBeforeSaving(
    String appName, String userId, String sessionId, String filename) {
  try {
    return readVersions(appName, userId, sessionId, filename);
  } catch (StorageException e) {
    throw new VerifyException("Failed to list artifact versions from GCS", e);
  }
}
```

`listVersions` has three callers besides the save path — `loadArtifact`, `deleteArtifact`, and
`ArtifactController` in the `dev` module — so removing the catch outright would change
all of them as well. Leaving it in place keeps the diff to the one caller that overwrites stored
data: those three behave exactly as before, and `listVersions_storageException_returnsEmptyList`
passes untouched. That is a scoping choice, not a claim that the empty list is right for them; the
issue lists them as related and uncovered. If you would rather the catch went away and those callers
moved with it, say so and I will make that change instead.

| File | Change |
|---|---|
| `core/src/main/java/…/artifacts/GcsArtifactService.java` | the four pieces above, `+81/-18` |
| `core/src/test/java/…/artifacts/GcsArtifactServiceTest.java` | 2 tests |

### Testing Plan

**Unit Tests:**

`GcsArtifactServiceTest`: **Tests run: 26, Failures: 0, Errors: 0, Skipped: 0.** Both new tests were
confirmed failing on unmodified `main` by reverting `GcsArtifactService.java` alone and re-running —
`Tests run: 26, Failures: 2`, exactly these two:

| Test | Asserts | On `main` |
|---|---|---|
| `save_listStorageException_propagates` | the save raises `VerifyException`, with `hasCauseThat().isInstanceOf(StorageException.class)` so the underlying failure stays visible | ❌ fails |
| `save_listStorageException_doesNotWrite` | `verify(mockStorage, never()).create(...)` | ❌ fails |

Both also `verify(mockStorage).list(...)`, so neither can pass by failing earlier for an unrelated
reason.

- [x] Run against a real GCS bucket, two arms. Both save `PAYLOAD-A` through a fully-permissioned
  client, then save `PAYLOAD-B` to the same filename — arm 1 through an identity denied
  `storage.objects.list`, arm 2 through a client that can list. The bucket is inspected afterwards
  with a privileged credential, since the restricted identity cannot list.

| | before | after |
|---|---|---|
| save 1, healthy client | v0 = `PAYLOAD-A` | v0 = `PAYLOAD-A` |
| save 2, client under test | reports v0, success | raises; the cause names `storage.objects.list` |
| bucket after | 1 object = `PAYLOAD-B` | 1 object = `PAYLOAD-A` |
| `loadArtifact(0)` | `PAYLOAD-B` | `PAYLOAD-A` |

The control arm is unchanged by the fix in both runs — versions `[0, 1]`, both objects kept — so the
two arms differ only in which client performs the second save.

### Checklist

### DIFF
--- a/core/src/main/java/com/google/adk/artifacts/GcsArtifactService.java
+++ b/core/src/main/java/com/google/adk/artifacts/GcsArtifactService.java
@@ -254,25 +254,10 @@ public final class GcsArtifactService implements BaseArtifactService {
       String appName, String userId, String sessionId, String filename) {
     return Single.fromCallable(
         () -> {
-          String prefix = getBlobPrefix(appName, userId, sessionId, filename);
           try {
-            return Streams.stream(
-                    storageClient.list(bucketName, BlobListOption.prefix(prefix)).iterateAll())
-                .map(Blob::getName)
-                .map(
-                    name -> {
-                      int versionDelimiterIndex = name.lastIndexOf('/');
-                      return versionDelimiterIndex != -1
-                              && versionDelimiterIndex < name.length() - 1
-                          ? Optional.of(name.substring(versionDelimiterIndex + 1))
-                          : Optional.<String>empty();
-                    })
-                .flatMap(Optional::stream)
-                .map(Integer::parseInt)
-                .sorted()
-                .collect(ImmutableList.toImmutableList());
+            return readVersions(appName, userId, sessionId, filename);
           } catch (StorageException e) {
-            return ImmutableList.of();
+            return ImmutableList.<Integer>of();
           }
         });
   }
@@ -314,7 +299,7 @@ public final class GcsArtifactService implements BaseArtifactService {
 
   private Single<SaveResult> saveArtifactAndReturnBlob(
       String appName, String userId, String sessionId, String filename, Part artifact) {
-    return listVersions(appName, userId, sessionId, filename)
+    return Single.fromCallable(() -> versionsBeforeSaving(appName, userId, sessionId, filename))
         .map(versions -> versions.isEmpty() ? 0 : max(versions) + 1)
         .flatMap(
             nextVersion ->
@@ -350,5 +335,65 @@ public final class GcsArtifactService implements BaseArtifactService {
                         throw new VerifyException("Failed to save artifact to GCS", e);
                       }
                     }));
+  }
+
+  /**
+   * Reads the versions stored for an artifact, letting a storage failure propagate.
+   *
+   * <p>An empty result from this method therefore means the artifact has no versions, never that
+   * the listing could not be performed. Each caller decides what a failure means for it: {@link
+   * #listVersions} reports it as "no versions", while {@link #saveArtifactAndReturnBlob} must not,
+   * because it derives the next version number from the result and would otherwise write over an
+   * object that already exists.
+   *
+   * @param appName Application name.
+   * @param userId User ID.
+   * @param sessionId Session ID.
+   * @param filename Artifact filename.
+   * @return sorted version numbers found under the artifact's prefix.
+   * @throws StorageException if the listing could not be performed.
+   */
+  private ImmutableList<Integer> readVersions(
+      String appName, String userId, String sessionId, String filename) {
+    String prefix = getBlobPrefix(appName, userId, sessionId, filename);
+    return Streams.stream(
+            storageClient.list(bucketName, BlobListOption.prefix(prefix)).iterateAll())
+        .map(Blob::getName)
+        .map(
+            name -> {
+              int versionDelimiterIndex = name.lastIndexOf('/');
+              return versionDelimiterIndex != -1 && versionDelimiterIndex < name.length() - 1
+                  ? Optional.of(name.substring(versionDelimiterIndex + 1))
+                  : Optional.<String>empty();
+            })
+        .flatMap(Optional::stream)
+        .map(Integer::parseInt)
+        .sorted()
+        .collect(ImmutableList.toImmutableList());
+  }
+
+  /**
+   * Reads the versions that already exist, for a save that is about to derive the next version
+   * number from them.
+   *
+   * <p>The write that follows carries no precondition, so treating a failed listing as "no
+   * versions" would compute version 0 and replace whatever is already stored under that name, while
+   * reporting success to the caller. Surfacing the failure instead is the same choice {@link
+   * #listArtifactKeys} and {@link #deleteArtifact} already make for the same operation.
+   *
+   * @param appName Application name.
+   * @param userId User ID.
+   * @param sessionId Session ID.
+   * @param filename Artifact filename.
+   * @return sorted version numbers already stored for the artifact.
+   * @throws VerifyException if the versions could not be listed.
+   */
+  private ImmutableList<Integer> versionsBeforeSaving(
+      String appName, String userId, String sessionId, String filename) {
+    try {
+      return readVersions(appName, userId, sessionId, filename);
+    } catch (StorageException e) {
+      throw new VerifyException("Failed to list artifact versions from GCS", e);
+    }
   }
 }
--- a/core/src/test/java/com/google/adk/artifacts/GcsArtifactServiceTest.java
+++ b/core/src/test/java/com/google/adk/artifacts/GcsArtifactServiceTest.java
@@ -20,6 +20,7 @@ import static org.junit.Assert.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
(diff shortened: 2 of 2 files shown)


hemasekhar-p:
Hi @svetanis, thank you for your contribution! We appreciate you taking the time to submit this pull request. I noticed that the branch is currently out of sync with the base branch. could you please resolve the merge conflicts so we can proceed with the review?

svetanis:
Hi @hemasekhar-p, thanks for taking a look.

I've rebased the branch onto the latest main (1929be77) — it's now up to date with the base branch and still a single commit.

There were no merge conflicts to resolve: the branch was simply behind main. It touches only GcsArtifactService.java and GcsArtifactServiceTest.java, and neither file was modified on main since the branch point — the two intervening commits touched RemoteA2AAgent and pom.xml.

The diff is unchanged from the original submission. Ready for review whenever it's convenient.

hemasekhar-p:
@svetanis, Thank you for the update. Currently this PR is under review by our team, we will keep you posted if any additional information is required. thank you.

sherryfox (approved):

PR #1427, opened 2026-08-12
fix: preserve raw tool parameter schemas in chat completions

### LINKED ISSUE #1426: ChatCompletionsRequest drops MCP tool parameter schemas from parametersJsonSchema
## 🔴 Required Information

**Describe the Bug:**

ADK's native `McpToolset` is unusable with the native chat-completions connector for MCP tools that take parameters.

`AbstractMcpTool` declares MCP tool input schemas through `FunctionDeclaration.parametersJsonSchema(...)`, but `com.google.adk.models.chat.ChatCompletionsRequest` reads only `FunctionDeclaration.parameters()`.

When only the raw JSON schema is populated, the chat-completions path falls back to an empty parameter schema:

```json
{
  "type": "object",
  "properties": {}
}
```

As a result, the model receives a parameterless tool declaration for an MCP tool that actually requires arguments and commonly calls the tool with `{}`.

This appears to be an incompatibility between two first-party `core/` ADK components rather than a consumer configuration issue.

**Steps to Reproduce:**

1. Create an MCP tool with a required parameter, for example:

   ```json
   {
     "type": "object",
     "properties": {
       "jobId": {
         "type": "string"
       }
     },
     "required": ["jobId"]
   }
   ```

2. Expose the MCP tool through ADK's native `McpToolset` / `AbstractMcpTool`.

3. Use the tool with a model backed by `ChatCompletionsHttpClient`.

4. Inspect the outbound `/chat/completions` request.

5. Observe that the generated tool declaration contains an empty parameter schema instead of the MCP tool's JSON schema.

**Expected Behavior:**

The outbound OpenAI-compatible tool declaration should preserve the MCP tool's parameter schema, including its properties and required fields:

```json
{
  "type": "function",
  "function": {
    "name": "analyze_premerge_failures_by_job",
    "parameters": {
      "type": "object",
      "properties": {
        "jobId": {
          "type": "string"
        }
      },
      "required": ["jobId"]
    }
  }
}
```

The model should therefore be able to identify `jobId` as a required argument and provide it when calling the tool.

**Observed Behavior:**

The outbound declaration contains an empty schema:

```json
{
  "type": "function",
  "function": {
    "name": "analyze_premerge_failures_by_job",
    "parameters": {
      "type": "object",
      "properties": {}
    }
  }
}
```

The model cannot see that `jobId` exists and commonly emits an empty argument object:

```json
{}
```

This prevents parameterized MCP tools from being used correctly through the chat-completions connector.

---

## 🟡 Optional Information

**Regression:**

No known regression has been established. The issue is observed with ADK `1.7.1`.

**Logs:**

The relevant evidence is the outbound `/chat/completions` request. The tool declaration contains the following schema instead of the MCP-provided schema:

```json
{
  "type": "object",
  "properties": {}
}
```

**Screenshots / Video:**

N/A.

**Additional Context:**

`FunctionDeclaration` supports two representations for function parameters:

* typed `parameters(Schema)`
* raw `parametersJsonSchema(...)`

`AbstractMcpTool` populates the raw `parametersJsonSchema` form for MCP tool input schemas. `ChatCompletionsRequest` currently consumes only the typed `parameters(Schema)` form.

We also observed the same raw-only declaration shape through an application-side MCP adapter, but the native `McpToolset` reproduction demonstrates that the issue does not depend on that adapter.

A closely related fix already exists for the Claude connector in PR #1342 / commit `760c8da`. That fix forwards the raw schema directly, preserving JSON Schema constructs such as `$ref` and `$defs`.

OpenAI-compatible chat-completions APIs also accept raw JSON Schema, so forwarding `parametersJsonSchema` directly appears preferable to converting it into the narrower typed `Schema` representation.

**Impact:**

Any parameterized MCP tool is affected when used through the chat-completions connector. In our case, this blocks:

* a tool accepting `jobId`
* parameterized architecture-analysis MCP tools

Zero-argument tools continue to work because their empty schema is valid.

**Temporary Workaround:**

Until ADK is fixed, we materialize the raw schemas into typed `parameters(Schema)` in our `BaseLlm` implementation before delegating to `ChatCompletionsHttpClient`.

This workaround is connector-local and is covered by a wire-level regression test.

A minimal reproduction should consist of:

```java
// Create an MCP tool whose input schema contains a required parameter,
// expose it through the native McpToolset, and invoke it through
// ChatCompletionsHttpClient.
//
// The outbound /chat/completions request contains:
// {
//   "type": "object",
//   "properties": {}
// }
// instead of the MCP tool's declared schema.
```

The important distinction is that the MCP tool declaration populates `parametersJsonSchema(...)`, while `ChatCompletionsRequest` currently reads `parameters()`.

* Always (100%)

dangzitou:
Fixes #1426

## Summary

- use `FunctionDeclaration.parametersJsonSchema` when building OpenAI-compatible tool declarations
- preserve the complete raw JSON Schema, including `required`, `$ref`, and `$defs`
- retain the existing typed `parameters` conversion and zero-argument fallback

This mirrors the raw-schema fallback already used by the Claude connector in #1342.

## Reproduction

Before the fix, the focused regression test was run three times. Every run failed with the same result: the output contained only an empty `properties` map, while `jobId`, `required`, and `$defs` were missing.

After the fix, `ChatCompletionsRequestTest` passes and the emitted tool schema equals the MCP-style raw schema.

## Validation

- `JAVA_HOME=/opt/homebrew/opt/openjdk@17/libexec/openjdk.jdk/Contents/Home ./mvnw -pl core -Dtest=ChatCompletionsRequestTest test`
- `JAVA_HOME=/opt/homebrew/opt/openjdk@17/libexec/openjdk.jdk/Contents/Home ./mvnw -pl core test`
- main test set: 1,769 tests, 0 failures, 0 errors, 24 skipped
- all additional Surefire profile tests passed; Maven finished with `BUILD SUCCESS`
- Google Java Format ran with 0 remaining reformats

No screenshot is needed because the regression test asserts the exact outbound tool schema object.

### DIFF
--- a/core/src/main/java/com/google/adk/models/chat/ChatCompletionsRequest.java
+++ b/core/src/main/java/com/google/adk/models/chat/ChatCompletionsRequest.java
@@ -603,7 +603,11 @@ public final class ChatCompletionsRequest {
             FunctionDefinition def = new FunctionDefinition();
             def.name = fd.name().orElse("");
             def.description = fd.description().orElse("");
-            if (fd.parameters().isPresent()) {
+            if (fd.parametersJsonSchema().isPresent()) {
+              def.parameters =
+                  objectMapper.convertValue(
+                      fd.parametersJsonSchema().get(), new TypeReference<Map<String, Object>>() {});
+            } else if (fd.parameters().isPresent()) {
               def.parameters =
                   objectMapper.convertValue(
                       fd.parameters().get(), new TypeReference<Map<String, Object>>() {});
--- a/core/src/main/java/com/google/adk/models/chat/ChatCompletionsRequest.java
+++ b/core/src/main/java/com/google/adk/models/chat/ChatCompletionsRequest.java
@@ -607,6 +607,9 @@ public final class ChatCompletionsRequest {
               def.parameters =
                   objectMapper.convertValue(
                       fd.parametersJsonSchema().get(), new TypeReference<Map<String, Object>>() {});
+              if (!def.parameters.containsKey("properties")) {
+                def.parameters.put("properties", ImmutableMap.of());
+              }
             } else if (fd.parameters().isPresent()) {
               def.parameters =
                   objectMapper.convertValue(
--- a/core/src/test/java/com/google/adk/models/chat/ChatCompletionsRequestTest.java
+++ b/core/src/test/java/com/google/adk/models/chat/ChatCompletionsRequestTest.java
@@ -37,6 +37,7 @@ import com.google.genai.types.Part;
 import com.google.genai.types.Schema;
 import com.google.genai.types.Tool;
 import com.google.genai.types.ToolConfig;
+import io.modelcontextprotocol.spec.McpSchema.JsonSchema;
 import java.util.AbstractMap;
 import java.util.Base64;
 import java.util.List;
@@ -645,6 +646,34 @@ public final class ChatCompletionsRequestTest {
 
     assertThat(request.tools).hasSize(1);
     assertThat(request.tools.get(0).function.parameters).isEqualTo(jsonSchema);
+  }
+
+  @Test
+  public void testFromLlmRequest_withZeroArgumentParametersJsonSchema() throws Exception {
+    JsonSchema jsonSchema = new JsonSchema("object", null, null, null, null, null);
+    FunctionDeclaration function =
+        FunctionDeclaration.builder()
+            .name("zero_argument_tool")
+            .parametersJsonSchema(jsonSchema)
+            .build();
+    Tool tool = Tool.builder().functionDeclarations(ImmutableList.of(function)).build();
+    GenerateContentConfig config =
+        GenerateContentConfig.builder().tools(ImmutableList.of(tool)).build();
+    LlmRequest llmRequest =
+        LlmRequest.builder()
+            .model("openai-compatible-model")
+            .config(config)
+            .contents(ImmutableList.of())
+            .build();
+
+    ChatCompletionsRequest request = ChatCompletionsRequest.fromLlmRequest(llmRequest, false);
+
+    assertThat(request.tools).hasSize(1);
+    Map<String, Object> params = (Map<String, Object>) request.tools.get(0).function.parameters;
+    assertThat(params.get("type")).isEqualTo("object");
+    @SuppressWarnings("unchecked")
+    Map<String, Object> props = (Map<String, Object>) params.get("properties");
+    assertThat(props).isEmpty();
   }
 
   @Test
--- a/core/src/test/java/com/google/adk/models/chat/ChatCompletionsRequestTest.java
+++ b/core/src/test/java/com/google/adk/models/chat/ChatCompletionsRequestTest.java
@@ -614,6 +614,40 @@ public final class ChatCompletionsRequestTest {
   }
 
   @Test
+  public void testFromLlmRequest_withParametersJsonSchema() throws Exception {
+    Map<String, Object> jsonSchema =
+        ImmutableMap.of(
+            "type",
+            "object",
+            "properties",
+            ImmutableMap.of("jobId", ImmutableMap.of("$ref", "#/$defs/jobId")),
+            "required",
+            ImmutableList.of("jobId"),
+            "$defs",
+            ImmutableMap.of("jobId", ImmutableMap.of("type", "string")));
+    FunctionDeclaration function =
+        FunctionDeclaration.builder()
+            .name("analyze_premerge_failures_by_job")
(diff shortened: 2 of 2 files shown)


hemasekhar-p:
Hi @dangzitou, thank you for your contribution! We really appreciate you taking the time to submit this pull request. As this is your first contribution to the project, please complete the Google CLA and feel free to mark this PR as ready for review so we can proceed with the next steps.

hemasekhar-p:
@dangzitou Thank you for the update. This PR is currently under review by our team. We will reach out if we require any additional information. Thank you!

core/src/test/java/com/google/adk/models/chat/ChatCompletionsRequestTest.java
```diff
@@ -613,6 +613,40 @@ public void testFromLlmRequest_withAbsentParameters() throws Exception {
...
+        LlmRequest.builder()
+            .model("openai-compatible-model")
+            .config(config)
+            .contents(ImmutableList.of())
+            .build();
+
+    ChatCompletionsRequest request = ChatCompletionsRequest.fromLlmRequest(llmRequest, false);
+
+    assertThat(request.tools).hasSize(1);
+    assertThat(request.tools.get(0).function.parameters).isEqualTo(jsonSchema);
```
sherryfox: The new test builds the schema as an `ImmutableMap`, but the bug in #1426 arrives as an `McpSchema.JsonSchema` record from `AbstractMcpTool.declaration()` — a different object graph through `convertValue`. I confirmed the record path works, so this isn't a correctness worry; a test shaped like the real producer would just guard the actual regression, and would have surfaced the zero-arg case above.

core/src/main/java/com/google/adk/models/chat/ChatCompletionsRequest.java
```diff
@@ -603,7 +603,11 @@ private static void handleTools(GenerateContentConfig config, ChatCompletionsReq
             FunctionDefinition def = new FunctionDefinition();
             def.name = fd.name().orElse("");
             def.description = fd.description().orElse("");
-            if (fd.parameters().isPresent()) {
+            if (fd.parametersJsonSchema().isPresent()) {
```
sherryfox, line 606: Zero-argument tools lose their `properties: {}` here. This branch returns before the else, which is the only place that adds it:
```
raw {"type":"object"}  →  main: {"type":"object","properties":{}}  →  now: {"type":"object"}
raw {}                 →  main: {"type":"object","properties":{}}  →  now: {}
```
The **else** explains why it matters: "OpenAI-compatible APIs (like Groq) strictly require the parameters object to exist."

Could you add the same padding here — if the converted map has no `properties` key, set it to `{}` before assigning? Schemas that already have one (like the `$ref`/`$defs` case you're fixing) should pass through untouched.

sherryfox (changes requested):

dangzitou:
Addressed the review feedback from @sherryfox:

- Added the missing `properties: {}` padding when a raw `parametersJsonSchema` has no properties (including the MCP `McpSchema.JsonSchema` record shape).
- Added `testFromLlmRequest_withZeroArgumentParametersJsonSchema` to cover the zero-argument MCP path.
- Re-ran `fmt:format` and `./mvnw -pl core -Dtest=ChatCompletionsRequestTest test`: 31 tests passed, BUILD SUCCESS.

core/src/test/java/com/google/adk/models/chat/ChatCompletionsRequestTest.java
dangzitou: Addressed in ce70332: the regression test now uses McpSchema.JsonSchema, matching the object produced by AbstractMcpTool.declaration(), while asserting preservation of properties, required, $ref, and $defs.

core/src/main/java/com/google/adk/models/chat/ChatCompletionsRequest.java
dangzitou: Addressed in 2fd0612 and covered by ce70332: converted raw schemas are padded with properties: {} when absent, including the zero-argument McpSchema.JsonSchema path. ChatCompletionsRequestTest passes.

dangzitou:
Addressed both requested changes: zero-argument raw schemas retain properties: {}, and the regression now uses the actual McpSchema.JsonSchema producer shape. Both review threads are resolved and ChatCompletionsRequestTest passes. @sherryfox, ready for re-review.

sherryfox:
Thanks @dangzitou!

One mechanical thing: Can you squash the commits since more than 1 on a PR fails our CI atm.

PR #1432, opened 2026-08-15
feat(plugins): port DebugLoggingPlugin from adk-python

### LINKED ISSUE #1431: [FEATURE] Port DebugLoggingPlugin from adk-python

When an agent misbehaves, adk-java has no way to produce a record of what happened that you can read
afterwards, diff against another run, or attach to a bug report. What was sent to the model and what
came back, each tool call and its result, the events the runner yielded  - all of it goes out as log
lines, mixed in with everything else the process is printing. Nothing gathers one invocation into one
place.

`adk-python` closed that gap by shipping `DebugLoggingPlugin`, which writes one YAML document per
invocation to a file. `adk-java` has no equivalent.

`DebugLoggingPlugin` in `com.google.adk.plugins.debuglogging`, registered on a `Runner` like any
other plugin, behaving as adk-python's does:

- record the user message, every LLM request and response, every tool call, result and error, every
  event the runner yields, and — optionally — a snapshot of session state as the invocation ends;
- write **one YAML document per invocation**, appended to a file, so a session's invocations
  accumulate in one readable artifact that can be attached to a bug report;
- **curate rather than dump**, as upstream does: tool declarations recorded as names only, inline
  data as its mime type with the bytes omitted, grounding metadata as a boolean, requested auth
  configs as a count;
- switch the session-state snapshot and full system instructions on or off per instance.

No existing class changes and no existing API changes. All twelve callbacks it needs are already
declared on `Plugin`, and every one of them is observe-only here: eleven return `Maybe.empty()`, and
`afterRunCallback` returns a `Completable` that only writes the file. None of them returns a value
that alters the run.

### Impact on your work

When an agent does something unexpected, there is nothing to save and nothing to hand to someone
else. Not blocking, and there is no timeline — this is a parity gap, not an outage.

### Willingness to contribute

Yes. A PR follows immediately after this issue: one new public plugin class plus supporting types in
a `plugins.debuglogging` subpackage, with tests. No existing file is modified, and no new dependency
is added — `jackson-dataformat-yaml` is already a compile-scope dependency of `core`.

---

**`LoggingPlugin`, which adk-java already has.** It logs prose lines to slf4j while the run happens,
truncating what it prints. That is useful for watching a run go by, but what it produces cannot be
re-read as data, diffed between two runs, or attached to an issue. Complementary, not a replacement —
the same hooks, a different output.

**ADK's tracing/telemetry.** Spans answer "where did the time go" and are read in a tracing backend.
A debug trace answers "what was said, in order", is read in a text editor, and needs no backend.

**`ReplayPlugin` and the `recordings` package in `google-adk-dev`.** Closer in spirit, but it reads
recordings rather than writing them, and it targets deterministic playback rather than a document a
person reads. It also ships in a different artifact from `core`.

**Write the plugin in application code.** It works — the behavior above was built using public API
only. But it is a few hundred lines of serialization curation per application, it drifts from
upstream separately in each copy, and the parts that matter most (what to omit) are exactly the
parts each copy will get wrong differently.

Registration needs no new API:

```java
Runner runner =
    new InMemoryRunner(
        agent, "my-app", ImmutableList.of(new DebugLoggingPlugin(Path.of("adk_debug.yaml"))));
```

All twelve hooks upstream overrides are already declared on `Plugin`, with compatible shapes - 
for example, the two that carry the model exchange, and the one that does the writing:

```java
default Maybe<LlmResponse> beforeModelCallback(
    CallbackContext callbackContext, LlmRequest.Builder llmRequest)

default Maybe<LlmResponse> afterModelCallback(
    CallbackContext callbackContext, LlmResponse llmResponse)

// returns Completable, so the file write need not block the caller
default Completable afterRunCallback(InvocationContext invocationContext)
```

The class is written and tested; the PR carries it.

svetanis:

- Closes: [#1431](https://github.com/google/adk-java/issues/1431)

**Problem:**

When an agent misbehaves, adk-java has no way to capture what happened in a form you can read
afterwards or attach to a bug report. The contents sent to the model, what came back, each tool call
and its result, the events the runner yielded — all of it is observable only while the run is
happening, and nothing produces a structured record. adk-python ships `DebugLoggingPlugin` for
exactly this; adk-java has no equivalent.

**Solution:**

A port of that plugin. New files in a new `com.google.adk.plugins.debuglogging` subpackage, **no
existing file modified**, no existing API changed, and **no new dependency** —
`jackson-dataformat-yaml` is already a compile-scope dependency of `core`. Registration uses the
surface that is already there:

```java
Runner runner =
    new InMemoryRunner(
        agent, "my-app", ImmutableList.of(new DebugLoggingPlugin(Path.of("adk_debug.yaml"))));
```

Twelve new files, the same count `plugins/agentanalytics` already ships:

| New file(s) | Concern |
|---|---|
| `DebugLoggingPlugin.java` | the twelve hooks; each body records one entry and returns |
| `DebugYamlWriter.java` | append one YAML document per invocation, serialized by a lock, and how a mapper is configured for traces |
| `DebugEntry.java`, `InvocationDebugState.java`, `DebugTraceRecorder.java` | one entry, one invocation's accumulator, and the live-invocation map that timestamped entries are filed into |
| `TracePayload.java` (sealed, + 8 nested) | the closed set of shapes an entry's `data` can take |
| `ContentTrace.java` (nesting `PartTrace` and its 6 part kinds), `EventTrace.java`, `LlmRequestTrace.java`, `LlmResponseTrace.java`, `UsageTrace.java` | the curated shape of each ADK/genai type as it appears in a trace |
| `SafeSerializer.java` | arbitrary tool values made YAML-safe without ever throwing |

The PR adds exactly **one** public type, `DebugLoggingPlugin`. Everything else is package-private.

**All twelve hooks are observe-only** — eleven return `Maybe.empty()`, and `afterRunCallback` returns
a `Completable` that only writes the file. Failures inside are logged, not thrown.

**The trace curates rather than dumps**, which is why this is not a generic `ObjectMapper` over ADK's
types. Tool declarations are recorded as names only; inline data as `mime_type` with
`_data_omitted: true` and no component for the bytes at all; grounding metadata as
`has_grounding_metadata: true`; requested auth configs as a count. Each matches upstream.

**One difference from adk-python.** `onUserMessageCallback` fires before `beforeRunCallback`, and
upstream opens its per-invocation state only in the later hook, so its `user_message` entry is
dropped. Here either hook opens the invocation, so the user's message is recorded. A test and a demo
row pin it.

On the write path:

- **The write is off the calling thread but still ordered.** `afterRunCallback` returns
  `Completable`, so the write runs on `Schedulers.io()`; `Runner.runAgentWithUpdatedSession` joins it
  onto the event stream with `concatWith`, so a caller that drains `runAsync` has the file on disk
  when it returns.
- **The writer is `synchronized`.** Two invocations finishing together would otherwise interleave
  their documents into a file that no longer parses — a wide window on the JVM, with the write on
  `Schedulers.io()`.

Ported from `adk-python/src/google/adk/plugins/debug_logging_plugin.py` (572 lines).

### Testing Plan

**Unit Tests:**

```text
mvn -o -pl core test -Dtest='com.google.adk.plugins.debuglogging.*Test'
Tests run: 97, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

Ninety-seven tests, all new — these are new classes, so there is no before/after state to contrast.
They assert on **serialized output**, not on intermediate maps, and the writer and plugin tests
assert on the **re-parsed YAML**, because "the trace can be read back" is the actual promise:

| Test class | Pins |
|---|---|
| `EventTraceTest` (17) | all thirteen envelope components on their wire keys; an absent streaming flag omitted rather than written `false`; the actions block only when non-empty; auth configs as a count; the live action maps copied |
| `LlmExchangeTraceTest` (15) | request and response shape; tool names only; `includeSystemInstruction=false` records presence, not text; response schema and grounding metadata each reduced to a boolean |
| `SafeSerializerTest` (12) | an arbitrary tool value never throws; nulls never reach an immutable collection; cycles are cut instead of overflowing the stack |
| `ContentTraceTest` (12) | all six part kinds on their wire keys; blank text and empty parts dropped; inline data keeps its mime type, never its bytes |
| `TracePayloadTest` (12) | the eight payload shapes on their wire keys; `data: {}` for markers; `args: {}` kept when a tool takes none |
| `DebugLoggingPluginTest` (9) | all twelve hooks record, in the order a real run fires them; the `Maybe` hooks return empty; the user message survives arriving before the run opens; both config switches; the state is released after the write |
| `DebugTraceRecorderTest` (9) | entry field order; an entry for an unknown invocation is dropped, not thrown; `entries()` is a snapshot, not the live queue; a session without app name or user id omits those keys |
| `DebugYamlWriterTest` (8) | two invocations, two documents, both re-parse; one `---` each; all four scalar-lookalikes survive as strings; multi-line text as a literal block; an unwritable path logs instead of throwing; eight threads appending at once still yield eight intact documents |
| `UsageTraceTest` (3) | three token counts for an event, four for a response — the cached-content count present in one and absent in the other |

One real invocation against `gemini-2.5-flash` through an `InMemoryRunner`, with one tool. Two model
calls, 15 entries, one document — excerpt:

```yaml
---
invocation_id: e-3b988d1a-5abc-4345-b1c1-acd6c5c84516
session_id: 339cf111-94ac-4014-b737-e3c4250f81d1
app_name: f15-debug-logging
user_id: demo-user
start_time: 2026-08-14T13:55:53.934
entries:
- timestamp: 2026-08-14T13:55:53.941
  entry_type: user_message
  invocation_id: e-3b988d1a-5abc-4345-b1c1-acd6c5c84516
  data:
    content:
      role: user
      parts:
      - text: "Where is order 42? Use the lookupOrder tool, then tell me in one sentence."
- timestamp: 2026-08-14T13:55:54.000
  entry_type: llm_request
  invocation_id: e-3b988d1a-5abc-4345-b1c1-acd6c5c84516
  agent_name: debug-logging-demo-agent
  data:
    model: gemini-2.5-flash
    content_count: 1
    contents:
    - role: user
      parts:
      - text: "Where is order 42? Use the lookupOrder tool, then tell me in one sentence."
    tools:
    - lookupOrder
- timestamp: 2026-08-14T13:55:55.547
  entry_type: tool_call
  invocation_id: e-3b988d1a-5abc-4345-b1c1-acd6c5c84516
  agent_name: debug-logging-demo-agent
  data:
    tool_name: lookupOrder
    function_call_id: adk-5f689fbe-8bbb-465d-b516-6303c70c2adf
    args:
      orderId: "42"
- timestamp: 2026-08-14T13:55:55.554
  entry_type: tool_response
  invocation_id: e-3b988d1a-5abc-4345-b1c1-acd6c5c84516
  agent_name: debug-logging-demo-agent
  data:
    tool_name: lookupOrder
    function_call_id: adk-5f689fbe-8bbb-465d-b516-6303c70c2adf
    result:
      orderId: "42"
      status: shipped
      carrier: DHL
```

Eight further scenarios were run offline against scripted models, covering the entry sequence, both
tool-failure shapes, a model error, append across two invocations, two invocations racing on one
file, a blob in the conversation, and the configuration switches.

### Checklist

### DIFF
--- a/core/src/main/java/com/google/adk/plugins/debuglogging/ContentTrace.java
+++ b/core/src/main/java/com/google/adk/plugins/debuglogging/ContentTrace.java
@@ -1,0 +1,201 @@
+/*
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.adk.plugins.debuglogging;
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static java.util.function.Predicate.not;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.genai.types.Blob;
+import com.google.genai.types.CodeExecutionResult;
+import com.google.genai.types.Content;
+import com.google.genai.types.ExecutableCode;
+import com.google.genai.types.FileData;
+import com.google.genai.types.FunctionCall;
+import com.google.genai.types.FunctionResponse;
+import com.google.genai.types.Part;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+/**
+ * One {@link Content} as it appears in the trace — port of adk-python's {@code _serialize_content}.
+ *
+ * <p>The factory requires a content; callers holding an {@code Optional<Content>} use {@link
+ * Optional#map} rather than passing null.
+ */
+@JsonInclude(JsonInclude.Include.NON_ABSENT)
+record ContentTrace(Optional<String> role, ImmutableList<PartTrace> parts) {
+
+  /** Always produced — an empty part list is still worth recording, as in the original. */
+  static ContentTrace from(Content content) {
+    return new ContentTrace(content.role(), traceParts(content));
+  }
+
+  private static ImmutableList<PartTrace> traceParts(Content content) {
+    return content.parts().orElseGet(ImmutableList::of).stream()
+        .map(PartTrace::from)
+        .flatMap(Optional::stream)
+        .collect(toImmutableList());
+  }
+
+  /**
+   * One {@link Part} as it appears in the trace — the declared shape of the per-part dict that
+   * adk-python builds in {@code _serialize_content}.
+   *
+   * <p>The record <em>is</em> the schema: the field names are the trace's keys, and a part kind
+   * that is not declared here cannot appear in the output.
+   */
+  @JsonInclude(JsonInclude.Include.NON_ABSENT)
+  record PartTrace(
+      Optional<String> text,
+      @JsonProperty("function_call") Optional<FunctionCallTrace> functionCall,
+      @JsonProperty("function_response") Optional<FunctionResponseTrace> functionResponse,
+      @JsonProperty("inline_data") Optional<InlineDataTrace> inlineData,
+      @JsonProperty("file_data") Optional<FileDataTrace> fileData,
+      @JsonProperty("code_execution_result") Optional<CodeExecutionResultTrace> codeExecutionResult,
+      @JsonProperty("executable_code") Optional<ExecutableCodeTrace> executableCode) {
+
+    /**
+     * The part's trace form, or empty when it carries nothing worth recording — adk-python's {@code
+     * if part_data:} guard.
+     */
+    static Optional<PartTrace> from(Part part) {
+      PartTrace trace =
+          new PartTrace(
+              part.text().filter(not(String::isEmpty)),
+              part.functionCall().map(FunctionCallTrace::from),
+              part.functionResponse().map(FunctionResponseTrace::from),
+              part.inlineData().map(InlineDataTrace::from),
+              part.fileData().map(FileDataTrace::from),
+              part.codeExecutionResult().map(CodeExecutionResultTrace::from),
+              part.executableCode().map(ExecutableCodeTrace::from));
+      return trace.isEmpty() ? Optional.empty() : Optional.of(trace);
+    }
+
+    /** Whether every declared field is absent, so the part would serialize to {@code {}}. */
+    @JsonIgnore
+    boolean isEmpty() {
+      return Stream.of(
+              text,
+              functionCall,
+              functionResponse,
+              inlineData,
+              fileData,
+              codeExecutionResult,
+              executableCode)
+          .allMatch(Optional::isEmpty);
+    }
+
+    /**
+     * A tool call the model requested.
+     *
+     * <p>{@code args} stays untyped because it genuinely is: a tool author chooses the argument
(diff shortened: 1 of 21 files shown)


hemasekhar-p:
Hi @svetanis, thank you for your contribution. We truly appreciate you taking the time to submit this PR. It is currently under review by our team and we will update you if any additional information is required. Thank you!

PR #1433, opened 2026-08-15
fix(core): retry invalid structured model responses

### LINKED ISSUE #1430: SetModelResponseTool: schema validation failure terminates the tool flow instead of returning feedback
### Describe the Bug

`SetModelResponseTool` currently propagates output-schema validation failures as tool execution errors.

In the current Java implementation, `SchemaUtils.validateMapOnSchema(...)` is executed inside `Single.fromCallable(...)`. When validation throws an `IllegalArgumentException`, the `Single` terminates with an error instead of returning validation feedback as a tool response.

Under the default tool execution flow, no validation-feedback `FunctionResponse` is returned to the model. As a result, the model cannot correct the arguments and call `set_model_response` again.

ADK Python now returns structured-output validation failures to the model as tool feedback and records successfully validated output separately. This appears to be a Java/Python parity gap.

There is also a related finalization concern: `OutputSchema.getStructuredModelResponse(...)` currently promotes a function response named `set_model_response` to the final structured response. Validation feedback must therefore be distinguishable from successfully validated output.

### Steps to Reproduce

Create an output schema containing a required field:

```java
Schema outputSchema =
    Schema.builder()
        .type("OBJECT")
        .properties(
            ImmutableMap.of(
                "field1",
                Schema.builder().type("STRING").build()))
        .required(ImmutableList.of("field1"))
        .build();
```

Create and invoke `SetModelResponseTool` without the required field:

```java
SetModelResponseTool tool =
    new SetModelResponseTool(outputSchema);

Map<String, Object> invalidArgs = ImmutableMap.of();

tool.runAsync(invalidArgs, null).blockingGet();
```

The call terminates with:

```text
java.lang.IllegalArgumentException:
Output args does not contain required field1
```

### Expected Behavior

For parity with ADK Python, an output-schema validation failure from the internal `set_model_response` tool should be returned to the model as validation feedback:

```text
LLM calls set_model_response
        ↓
schema validation fails
        ↓
validation feedback is returned to the model
        ↓
LLM corrects the arguments
        ↓
LLM calls set_model_response again
        ↓
validated result becomes the final response
```

An invalid `set_model_response` call must not become the final structured model response. Only successfully validated output should be finalized.

The exact Java implementation does not need to identify validation feedback solely through an `"error"` field, since a valid user-defined output schema could legitimately contain a field with that name.

### Observed Behavior

`SetModelResponseTool` currently performs:

```java
return Single.fromCallable(
    () -> {
      SchemaUtils.validateMapOnSchema(
          args, outputSchema, /* isInput= */ false);
      return args;
    });
```

The validation exception therefore becomes an RxJava error signal and propagates through the tool flow.

The existing upstream `SetModelResponseToolTest` explicitly verifies this behavior using `assertThrows(...)`, confirming that exception propagation is the current expected Java behavior.

Additionally, `OutputSchema.getStructuredModelResponse(...)` currently identifies the final structured response using the `set_model_response` function name. A change that only converts the exception into a Python-style feedback map could therefore incorrectly promote that feedback to the final response unless validated output is tracked or distinguished separately.

### Environment Details

- ADK Java version: commit `2b87d65d9704a61ff4668b8c9482a79fef9fe0d4`
- Java: Microsoft OpenJDK 17.0.15
- OS: Windows 11
- Test framework: JUnit 4

### Model Information

- Model: N/A
- The issue is reproducible deterministically at the tool level without a live model

---

### Regression

No known Java regression.

This appears to be a behavioral parity gap following the newer structured-output validation recovery behavior in ADK Python.

### Logs

```text
IllegalArgumentException: Output args does not contain required field1
    at com.google.adk.SchemaUtils.validateMapOnSchema(SchemaUtils.java:124)
    at com.google.adk.tools.SetModelResponseTool.lambda$runAsync$0(
        SetModelResponseTool.java:63)
```

### Additional Context

The corresponding Python behavior is present on `adk-python` main in:

https://github.com/google/adk-python/commit/b5b27cb074d00f2e6889098514d8f716bc1cdfad

The change originated from:

https://github.com/google/adk-python/pull/6368

The Python implementation catches structured-output validation failures and returns validation feedback to the model. Successfully validated output is recorded separately and is the only result promoted to the final structured response.

Relevant Java files:

- `core/src/main/java/com/google/adk/tools/SetModelResponseTool.java`
- `core/src/main/java/com/google/adk/flows/llmflows/OutputSchema.java`
- `core/src/test/java/com/google/adk/tools/SetModelResponseToolTest.java`

The ADK Java contribution guide states that ADK Java aims to align with ADK Python and leans on ADK Python as the source of truth for behavioral validation:

https://github.com/google/adk-java/blob/2b87d65d9704a61ff4668b8c9482a79fef9fe0d4/CONTRIBUTING.md#alignment-with-adk-python

Would maintainers agree that ADK Java should align with the current Python recovery behavior here?

### How often has this issue occurred?

Always (100%) when `SetModelResponseTool` receives arguments that fail output-schema validation.

BlueCatPro:

**1. Link to an existing issue:**

- Closes: #1430

**Problem:**

When `set_model_response` arguments fail output-schema validation, the validation exception prevents the model from receiving actionable feedback and retrying with a corrected structured response.

**Solution:**

- Return concise schema-validation feedback from `set_model_response` so the model can retry.
- Record successfully validated responses in `EventActions`.
- Promote only validated responses to the final structured model output.
- Remove the appended full schema from validation feedback to keep it concise.
- Add unit and Runner-level regression tests covering validation failure, feedback propagation, retry, and successful final output.

### Testing Plan

**Unit Tests:**

- [x] The unit tests relevant to this change pass locally.

Tests executed:

- `EventActionsTest`
- `OutputSchemaTest`
- `SetModelResponseToolTest`

Result:

- `BUILD SUCCESS`
- 27 tests
- 0 failures
- 0 errors

Not run against a live model. A Runner-level regression test using the scripted `TestLlm` verifies that validation feedback from the invalid call is included in the retry request and that only the corrected, validated response becomes the final structured output.

### Checklist

- [x] I have read the `CONTRIBUTING.md` document.
- [x] I have added tests that prove my fix is effective.
- [x] New and existing tests relevant to this change pass locally.
- [ ] I have manually tested the change end-to-end with a live model.
- [x] No dependent downstream changes are required.

### Additional context

This is my first open-source contribution and my first contribution to adk-java. I may have missed some repository conventions, so I would appreciate any guidance and am happy to make follow-up changes. Thank you for your time and review!

### DIFF
--- a/core/src/main/java/com/google/adk/flows/llmflows/OutputSchema.java
+++ b/core/src/main/java/com/google/adk/flows/llmflows/OutputSchema.java
@@ -78,18 +78,24 @@ public final class OutputSchema implements RequestProcessor {
   }
 
   /**
-   * Check if function response contains set_model_response and extract JSON.
+   * Extracts a successfully validated {@code set_model_response} result as JSON.
+   *
+   * <p>Only a result that passed output-schema validation (recorded on the event actions by {@link
+   * SetModelResponseTool}) is returned. Validation feedback sent back to the model is never
+   * promoted to the final structured response.
    *
    * @param functionResponseEvent The function response event to check.
-   * @return JSON response string if set_model_response was called, Optional.empty() otherwise.
+   * @return JSON response string if set_model_response succeeded, Optional.empty() otherwise.
    */
   public static Optional<String> getStructuredModelResponse(Event functionResponseEvent) {
     for (FunctionResponse funcResponse : functionResponseEvent.functionResponses()) {
       if (Objects.equals(funcResponse.name().orElse(""), SetModelResponseTool.NAME)) {
-        Object response = funcResponse.response();
-        // The tool returns the args map directly.
+        Optional<Object> validatedResponse = functionResponseEvent.actions().setModelResponse();
+        if (validatedResponse.isEmpty()) {
+          return Optional.empty();
+        }
         try {
-          return Optional.of(JsonBaseModel.getMapper().writeValueAsString(response));
+          return Optional.of(JsonBaseModel.getMapper().writeValueAsString(validatedResponse.get()));
         } catch (JsonProcessingException e) {
           logger.error("Failed to serialize set_model_response result", e);
           return Optional.empty();
--- a/core/src/main/java/com/google/adk/tools/SetModelResponseTool.java
+++ b/core/src/main/java/com/google/adk/tools/SetModelResponseTool.java
@@ -17,9 +17,13 @@
 package com.google.adk.tools;
 
 import com.google.adk.SchemaUtils;
+import com.google.common.collect.ImmutableMap;
 import com.google.genai.types.FunctionDeclaration;
 import com.google.genai.types.Schema;
 import io.reactivex.rxjava3.core.Single;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
@@ -32,6 +36,12 @@ import java.util.Optional;
  */
 public class SetModelResponseTool extends BaseTool {
   public static final String NAME = "set_model_response";
+
+  // Prefix of the SchemaUtils validation message after which the full schema is appended. Used to
+  // strip the schema dump from feedback on a best-effort basis; if SchemaUtils changes its wording
+  // the feedback simply stays unstripped. runAsync_unknownArg_feedbackOmitsSchemaDump pins the
+  // current format.
+  private static final String OUTPUT_SCHEMA_DUMP_MARKER = " does not match agent output schema: ";
 
   private final Schema outputSchema;
 
@@ -56,12 +66,63 @@ public class SetModelResponseTool extends BaseTool {
 
   @Override
   public Single<Map<String, Object>> runAsync(Map<String, Object> args, ToolContext toolContext) {
-    // This tool is a marker for the final response, it doesn't do anything but return its arguments
-    // which will be captured as the final result.
+    // Record validated responses on the event actions; return validation feedback so the model can
+    // retry.
     return Single.fromCallable(
         () -> {
-          SchemaUtils.validateMapOnSchema(args, outputSchema, /* isInput= */ false);
-          return args;
+          try {
+            SchemaUtils.validateMapOnSchema(args, outputSchema, /* isInput= */ false);
+          } catch (IllegalArgumentException e) {
+            return ImmutableMap.of(
+                "error",
+                "Validation Error found:\n"
+                    + sanitizeValidationMessage(e.getMessage())
+                    + "\nRecall the set_model_response function correctly, fix the errors, and"
+                    + " call it again with all required fields using the correct types.");
+          }
+          // Match Python's model_dump(exclude_none=True) for Java's map-shaped response.
+          Map<String, Object> validatedResponse = excludeNullFields(args);
+          toolContext.actions().setSetModelResponse(validatedResponse);
+          return validatedResponse;
         });
+  }
+
+  private static Map<String, Object> excludeNullFields(Map<String, Object> values) {
+    Map<String, Object> result = new LinkedHashMap<>();
+    for (Map.Entry<String, Object> entry : values.entrySet()) {
+      Object value = entry.getValue();
+      if (value != null) {
+        result.put(entry.getKey(), excludeNullFields(value));
+      }
+    }
+    return result;
+  }
+
+  @SuppressWarnings("unchecked")
+  private static Object excludeNullFields(Object value) {
+    if (value instanceof Map<?, ?>) {
+      return excludeNullFields((Map<String, Object>) value);
+    }
+    if (value instanceof List<?>) {
+      List<Object> result = new ArrayList<>();
+      for (Object item : (List<?>) value) {
+        result.add(excludeNullFields(item));
+      }
+      return result;
+    }
+    return value;
+  }
+
+  private static String sanitizeValidationMessage(String message) {
+    if (message == null) {
+      return "Arguments do not match the output schema.";
+    }
+    // The model already knows the schema from the tool declaration, so the appended schema dump is
+    // redundant in feedback.
+    int schemaDumpIndex = message.indexOf(OUTPUT_SCHEMA_DUMP_MARKER);
+    if (schemaDumpIndex >= 0) {
+      message = message.substring(0, schemaDumpIndex) + " does not match agent output schema.";
+    }
+    return message;
   }
 }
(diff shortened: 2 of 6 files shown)


hemasekhar-p:
Hi @BlueCatPro, thank you for your contribution! We appreciate you taking the time to submit this pull request. Currently this PR is under review by our team, we will keep you posted if any additional information is required. thank you.

core/src/main/java/com/google/adk/tools/SetModelResponseTool.java
```diff
@@ -56,12 +63,35 @@ public Optional<FunctionDeclaration> declaration() {
...
+            SchemaUtils.validateMapOnSchema(args, outputSchema, /* isInput= */ false);
+          } catch (IllegalArgumentException e) {
+            return ImmutableMap.of(
+                "error",
+                "Validation Error found:\n"
+                    + sanitizeValidationMessage(e.getMessage())
+                    + "\nRecall the set_model_response function correctly, fix the errors, and"
+                    + " call it again with all required fields using the correct types.");
+          }
+          toolContext.actions().setSetModelResponse(args);
```
sherryfox: Python records the validated and dumped object here, not the input:

```python
validated_response = self.output_schema.model_validate(args)
result = validated_response.model_dump(exclude_none=True)
...
tool_context.actions.set_model_response = result
```

 `SchemaUtils.validateMapOnSchema` returns void, so there's nothing to store here but args. Confirmed on this branch: `{"a":"x","b":null}` passes validation unchanged and lands in `setSetModelResponse` as-is, where Python would emit `{"a":"x"}`.

 Is the difference intended?

sherryfox (changes requested):
Hey @BlueCatPro, thanks for this — and for the issue write-up that preceded it.

I left one parity question about what gets stored so please take a look 🙏

core/src/main/java/com/google/adk/tools/SetModelResponseTool.java
BlueCatPro: Thanks for catching this, and sorry I missed this distinction. The difference was not intended.
I’ve updated `SetModelResponseTool` to create a copy after successful validation, omit null-valued fields to match the concrete `model_dump(exclude_none=True)` behavior you identified, and use the same result both as the tool return value and the value stored in `EventActions`.

core/src/main/java/com/google/adk/tools/SetModelResponseTool.java
```diff
@@ -56,12 +64,42 @@ public Optional<FunctionDeclaration> declaration() {
...
+            SchemaUtils.validateMapOnSchema(args, outputSchema, /* isInput= */ false);
+          } catch (IllegalArgumentException e) {
+            return ImmutableMap.of(
+                "error",
+                "Validation Error found:\n"
+                    + sanitizeValidationMessage(e.getMessage())
+                    + "\nRecall the set_model_response function correctly, fix the errors, and"
+                    + " call it again with all required fields using the correct types.");
+          }
+          // Match Python's model_dump(exclude_none=True) for Java's map-shaped response.
```
sherryfox, line 83: Thanks — this is what I meant, and using the same map for the return value and the actions is right.

One thing the loop does not cover: pydantic's `exclude_none=True` is recursive, and this filters only the top level. Same input both sides:

```
python  {'name':'x','addr':{'city':'NYC','zip':None}}  ->  {'name':'x','addr':{'city':'NYC'}}
here    {name=x, addr={city=NYC, zip=null}}
```

Narrow in practice — it needs a nested nullable property that the model actually sends as null, since `SchemaUtils.matchType` rejects a null on a non-nullable one before you get here. But the comment above says "Match Python's `model_dump(exclude_none=True)`" and the test is named `..._excludesNullFromReturnedAndRecordedResponse`, so either the filter should recurse or those two should say top-level only.

core/src/main/java/com/google/adk/flows/llmflows/OutputSchema.java
```diff
@@ -78,18 +78,24 @@ public Single<RequestProcessingResult> processRequest(
...
    * @param functionResponseEvent The function response event to check.
-   * @return JSON response string if set_model_response was called, Optional.empty() otherwise.
+   * @return JSON response string if set_model_response succeeded, Optional.empty() otherwise.
    */
   public static Optional<String> getStructuredModelResponse(Event functionResponseEvent) {
     for (FunctionResponse funcResponse : functionResponseEvent.functionResponses()) {
       if (Objects.equals(funcResponse.name().orElse(""), SetModelResponseTool.NAME)) {
-        Object response = funcResponse.response();
-        // The tool returns the args map directly.
+        Optional<Object> validatedResponse = functionResponseEvent.actions().setModelResponse();
```
sherryfox, line 93: This matches `_output_schema_processor.py` exactly — reading the recorded value rather than the raw function response is what stops feedback being promoted, so no change asked for here.

Worth a test though: an event whose `set_model_response` never succeeded should yield `Optional.empty()` from `getStructuredModelResponse`, so the "feedback is never the final output" property is pinned rather than implied.

sherryfox (changes requested):

core/src/main/java/com/google/adk/tools/SetModelResponseTool.java
BlueCatPro: Ah, you’re right — my first pass missed nested nulls. Sorry about that. Just pushed a recursive fix + tests for nested objects and objects inside arrays.

core/src/main/java/com/google/adk/flows/llmflows/OutputSchema.java
BlueCatPro: Thanks for the suggestion. `getStructuredModelResponse_withValidationFeedback_returnsEmpty` covers this case by constructing validation feedback without a recorded `setModelResponse` and verifying that `getStructuredModelResponse` returns `Optional.empty()`.

sherryfox (approved):

damianmomotgoogle:
Thanks for the contribution, and welcome! This is a port of the Python set_model_response retry behavior but actual implementation looks to have a bug.

SetModelResponseTool.java — excludeNullFields (the two overloads)
excludeNullFields navigates the Schema, but the Python reference does this
purely structurally via model_dump(exclude_none=True). The schema coupling
introduces a latent crash: SchemaUtils.matchType only touches items() inside its
element loop, so an ARRAY field declared without `items` plus an empty-list value
passes validation, but excludeNullFields calls schema.items().get()
unconditionally and throws NoSuchElementException.

BlueCatPro:
@damianmomotgoogle Thanks for catching this. Sorry I missed it. I've fixed the null filtering and added a test for this case.

wikaaaaa:
Hi @BlueCatPro, thanks for addressing the comments! Everything looks good now, but the branch has merge conflicts with main. Could you please sync/rebase and resolve the conflicts?

BlueCatPro:
@wikaaaaa Thanks! I've resolved the merge conflicts.
Please let me know if there's anything else I should address.

wikaaaaa (approved):

PR #1448, opened 2026-08-21
fix: preserve complete MCP call tool results

### LINKED ISSUE #1443: McpTool drops structuredContent and non-text content from CallToolResult
## 🔴 Required Information

**Describe the Bug:**

`AbstractMcpTool.wrapCallResult(...)` silently discards parts of a successful MCP `CallToolResult`.

When the result contains at least one `TextContent`, native `McpTool.runAsync(...)` returns only the legacy `text_output` map. It does not expose:

* `CallToolResult.isError` on successful results
* `CallToolResult.structuredContent`
* non-text `content` entries such as images or embedded resources
* the original ordered content list

This loss is irreversible for callers. An outer `BaseTool` decorator sees only the already-reduced map and cannot recover the discarded fields without bypassing native `McpTool` and invoking `McpSyncClient` itself.

**Steps to Reproduce:**

1. Use `com.google.adk:google-adk:1.7.1`.
2. Construct a native `McpToolset` over a streamable-HTTP MCP server.
3. Have the server advertise a zero-argument tool.
4. Return this successful `tools/call` result:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "content": [
      {"type": "text", "text": "first"},
      {"type": "text", "text": "second"},
      {
        "type": "image",
        "data": "aW1hZ2U=",
        "mimeType": "image/png"
      }
    ],
    "structuredContent": {"count": 2},
    "isError": false
  }
}
```

5. Call the discovered native tool through `BaseTool.runAsync(...)`.

**Expected Behavior:**

Native MCP tools should preserve the complete semantic `CallToolResult`, including structured content and non-text content, without breaking callers that consume the existing `text_output` field.

A compatible resolution could be either:

1. An additive envelope that retains `text_output` unchanged while also exposing `content`, `structuredContent`, `isError`, and metadata.
2. A supported result-mapper/adapter hook on `McpToolset` or `McpTool` that receives the raw `CallToolResult` before reduction.

**Observed Behavior:**

The returned map contains only:

```text
{text_output=[{text=first}, {text=second}]}
```

`structuredContent`, image data and MIME type, the original content parts, and explicit non-error state are absent.

The loss is silent. No exception or warning indicates that valid MCP response data was discarded.

**Environment Details:**

* ADK Library Version (see maven dependency): `1.7.1`
* MCP Java SDK resolved by ADK: `1.1.2`
* MCP transport: streamable HTTP
* Java: 21

## 🟡 Optional Information

**Regression:**

No

**Additional Context:**

PR #1078 changed `wrapCallResult` to serialize the complete `CallToolResult`, but PR #1090 reverted that change to preserve backward compatibility with `text_output`.

The requested fix is therefore not to remove `text_output`. It is to preserve compatibility without silently losing the remaining MCP result fields.

This blocks replacing a custom MCP bridge with native `McpToolset`: the custom implementation currently preserves text, `structuredContent`, non-text content parts, and error state, while native `McpTool` does not.

N/A. The issue can be reproduced using the MCP response above and a native `McpToolset` over a streamable-HTTP MCP server.

* Always (100%)

eyesofish:
## Summary

- preserve the complete successful MCP `CallToolResult`, including ordered content, structured content, metadata, and `isError`
- retain the legacy `text_output` field for backward compatibility
- cover mixed text/image and non-text-only results

## Testing

- `./mvnw fmt:format`
- `./mvnw -pl core -Dtest=AbstractMcpToolTest,ClaudeTest test`
- `./mvnw -pl core test` (1,751 tests, 0 failures)

Closes #1443

> Replaces #1446 (accidentally closed; GitHub blocks reopening after the head branch was force-pushed).

### DIFF
--- a/core/src/main/java/com/google/adk/tools/mcp/AbstractMcpTool.java
+++ b/core/src/main/java/com/google/adk/tools/mcp/AbstractMcpTool.java
@@ -131,8 +131,11 @@ public abstract class AbstractMcpTool<T> extends BaseTool {
       return ImmutableMap.of("error", errorMessage);
     }
 
+    Map<String, Object> resultMap =
+        objectMapper.convertValue(callResult, new TypeReference<Map<String, Object>>() {});
+
     if (contents == null || contents.isEmpty()) {
-      return ImmutableMap.of();
+      return resultMap;
     }
 
     List<String> textOutputs = new ArrayList<>();
@@ -145,11 +148,7 @@ public abstract class AbstractMcpTool<T> extends BaseTool {
     }
 
     if (textOutputs.isEmpty()) {
-      return ImmutableMap.of(
-          "error",
-          "Tool '" + mcpToolName + "' returned content that is not TextContent.",
-          "content_details",
-          contents.toString());
+      return resultMap;
     }
 
     List<Map<String, Object>> resultMaps = new ArrayList<>();
@@ -161,6 +160,7 @@ public abstract class AbstractMcpTool<T> extends BaseTool {
         resultMaps.add(ImmutableMap.of("text", textOutput));
       }
     }
-    return ImmutableMap.of("text_output", resultMaps);
+    resultMap.put("text_output", resultMaps);
+    return resultMap;
   }
 }
--- a/core/src/test/java/com/google/adk/tools/mcp/AbstractMcpToolTest.java
+++ b/core/src/test/java/com/google/adk/tools/mcp/AbstractMcpToolTest.java
@@ -25,6 +25,7 @@ import com.google.common.collect.ImmutableList;
 import io.modelcontextprotocol.client.McpSyncClient;
 import io.modelcontextprotocol.spec.McpSchema;
 import io.modelcontextprotocol.spec.McpSchema.CallToolResult;
+import io.modelcontextprotocol.spec.McpSchema.ImageContent;
 import io.modelcontextprotocol.spec.McpSchema.TextContent;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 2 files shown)


hemasekhar-p:
Hi @eyesofish, thank you for your contribution! We appreciate you taking the time to submit this pull request. Currently this PR is under review by our team we will keep you posted if any additional information is required. thank you.

core/src/main/java/com/google/adk/tools/mcp/AbstractMcpTool.java
```diff
@@ -131,8 +131,11 @@ protected static Map<String, Object> wrapCallResult(
       return ImmutableMap.of("error", errorMessage);
     }
 
+    Map<String, Object> resultMap =
+        objectMapper.convertValue(callResult, new TypeReference<Map<String, Object>>() {});
+
     if (contents == null || contents.isEmpty()) {
-      return ImmutableMap.of();
+      return resultMap;
```
sherryfox, line 138: This branch used to return `ImmutableMap.of()`; it now returns the full envelope which means that a consumer checking `result.isEmpty()` would silently stop matching.

That is a behaviour change and it is the one part of the diff with no test and no javadoc - can you possibly update this? The two new tests cover mixed content and non-text content; a third case here would close the gap.

core/src/main/java/com/google/adk/tools/mcp/AbstractMcpTool.java
```diff
@@ -161,6 +160,7 @@ protected static Map<String, Object> wrapCallResult(
         resultMaps.add(ImmutableMap.of("text", textOutput));
       }
     }
-    return ImmutableMap.of("text_output", resultMaps);
+    resultMap.put("text_output", resultMaps);
```
sherryfox, line 163: The duplication itself is fine — keeping text_output alongside content is the point of an additive migration, and it's what the issue asked for. WDYT of marking it `@Deprecated` and noting it in the doc, so the legacy half doesn't become permanent?

Measured overhead of carrying both, serialized into the function response sent to the model: 
```
payload            legacy-only   with envelope   overhead
"success"                   36              97     +169%
200 ch prose               229             483     +111%
2000 ch prose             2029            4083     +101%
JSON tool result            85             222     +161%
```
~2× tokens on every MCP tool call, indefinitely.

sherryfox (changes requested):

PR #1449, opened 2026-08-23
fix(artifacts): honor the "user:" namespace in InMemoryArtifactService

shoemoney:

None found.

**Problem:**

ADK has a `"user:"` filename-prefix convention: a filename starting with `user:` is scoped to the user across all their sessions, not just the session it was saved from.

`GcsArtifactService` implements this. `fileHasUserNamespace(filename)` (GcsArtifactService.java:65) checks the prefix, and `getBlobPrefix` (GcsArtifactService.java:78-84) branches on it to store at `appName/userId/user/filename/version`, deliberately leaving the sessionId segment out. This is exercised directly in `GcsArtifactServiceTest` (`save_userNamespace_savesCorrectly`, `load_userNamespace_loadsCorrectly`).

`InMemoryArtifactService` had no equivalent branch. Its `getArtifactsMap(appName, userId, sessionId)` always keyed by `appName -> userId -> sessionId -> filename`, and every public method (`saveArtifact`, `loadArtifact`, `listArtifactKeys`, `deleteArtifact`, `listVersions`) went through it. So if code called `CallbackContext.saveArtifact("user:config.json", part)` in one session, then `loadArtifact("user:config.json")` in a different session for the same app/user, it got `Maybe.empty()`. No exception, just silently missing data.

This matters because `InMemoryArtifactService` is the default backend used in local dev, tests, and the quickstarts. An app can be written and tested entirely against `InMemoryArtifactService`, rely on the documented cross-session behavior of `user:`-prefixed filenames, and only discover it never worked once it's pointed at `GcsArtifactService` in prod, or never discover it at all if it stays on `InMemoryArtifactService`.

**Solution:**

Mirror the GCS branch. `getArtifactsMap` now also takes the filename, using the same `fileHasUserNamespace` check and naming as `GcsArtifactService`, and routes user-namespaced filenames to a shared per-user bucket (`USER_NAMESPACE_SESSION_KEY`) instead of the caller's `sessionId`. `saveArtifact`, `loadArtifact`, `deleteArtifact`, and `listVersions` all pass the filename through unchanged. `listArtifactKeys` doesn't have a filename to branch on, so it now merges the session-scoped bucket with the shared user-namespace bucket, which is the same thing `GcsArtifactService#listArtifactKeys` does (it merges a session-prefix listing with a user-prefix listing).

No behavior changes for non-`user:`-prefixed filenames. `GcsArtifactService` is untouched.

### Testing Plan

**Unit Tests:**

Added to `InMemoryArtifactServiceTest`, mirroring the shape of the `GcsArtifactServiceTest` user-namespace cases:

- `save_userNamespace_visibleAcrossSessions` - save from session A, load from session B, same value comes back.
- `load_nonNamespacedFilename_notVisibleAcrossSessions` - regression guard. A non-prefixed filename saved in session A must still be invisible from session B. This is what proves the fix scopes only the `user:` case instead of making every artifact global.
- `listArtifactKeys_userNamespace_visibleAcrossSessions`
- `listVersions_userNamespace_visibleAcrossSessions`
- `deleteArtifact_userNamespace_removesAcrossSessions`

I ran `InMemoryArtifactServiceTest` against the unpatched source first to confirm these actually catch the bug:

```
Tests run: 8, Failures: 4, Errors: 0, Skipped: 0
InMemoryArtifactServiceTest.save_userNamespace_visibleAcrossSessions - FAILED
InMemoryArtifactServiceTest.listArtifactKeys_userNamespace_visibleAcrossSessions - FAILED
InMemoryArtifactServiceTest.listVersions_userNamespace_visibleAcrossSessions - FAILED
InMemoryArtifactServiceTest.deleteArtifact_userNamespace_removesAcrossSessions - FAILED
```

`load_nonNamespacedFilename_notVisibleAcrossSessions` and the three pre-existing tests passed in both states, as expected.

With the fix applied:

```
mvn -pl core -am test -Dtest=InMemoryArtifactServiceTest
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0
```

Also ran `GcsArtifactServiceTest` to confirm no regression there, since it's untouched:

```
mvn -pl core -am test -Dtest=GcsArtifactServiceTest
Tests run: 24, Failures: 0, Errors: 0, Skipped: 0
```

Not applicable, this is contained to `InMemoryArtifactService`'s in-process map and is fully covered by the unit tests above.

### Checklist

### Additional context

I checked the other two open PRs that touch artifacts, #1378 and #1412. Both are about `Runner`/plugin-level input-blob offload and don't touch `InMemoryArtifactService` or the user-namespace behavior, so there's no overlap with this change.

### DIFF
--- a/core/src/main/java/com/google/adk/artifacts/InMemoryArtifactService.java
+++ b/core/src/main/java/com/google/adk/artifacts/InMemoryArtifactService.java
@@ -81,7 +81,7 @@ public final class InMemoryArtifactService implements BaseArtifactService {
       String appName, String userId, String sessionId, String filename, @Nullable Integer version) {
     List<Part> versions =
         getArtifactsMap(appName, userId, sessionId, filename)
-            .computeIfAbsent(filename, unused -> new ArrayList<>());
+            .getOrDefault(filename, ImmutableList.of());
 
     if (versions.isEmpty()) {
       return Maybe.empty();
@@ -113,7 +113,7 @@ public final class InMemoryArtifactService implements BaseArtifactService {
     filenames.addAll(
         getArtifactsMapForSessionKey(appName, userId, USER_NAMESPACE_SESSION_KEY).keySet());
     return Single.just(
-        ListArtifactsResponse.builder().filenames(ImmutableList.copyOf(filenames)).build());
+        ListArtifactsResponse.builder().filenames(ImmutableList.sortedCopyOf(filenames)).build());
   }
 
   /**
@@ -138,7 +138,7 @@ public final class InMemoryArtifactService implements BaseArtifactService {
       String appName, String userId, String sessionId, String filename) {
     int size =
         getArtifactsMap(appName, userId, sessionId, filename)
-            .computeIfAbsent(filename, unused -> new ArrayList<>())
+            .getOrDefault(filename, ImmutableList.of())
             .size();
     if (size == 0) {
       return Single.just(ImmutableList.of());
--- a/core/src/main/java/com/google/adk/artifacts/InMemoryArtifactService.java
+++ b/core/src/main/java/com/google/adk/artifacts/InMemoryArtifactService.java
@@ -26,17 +26,34 @@ import io.reactivex.rxjava3.core.Maybe;
 import io.reactivex.rxjava3.core.Single;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.IntStream;
 import org.jspecify.annotations.Nullable;
 
 /** An in-memory implementation of the {@link BaseArtifactService}. */
 public final class InMemoryArtifactService implements BaseArtifactService {
+  // Mirrors the "user" path segment GcsArtifactService.getBlobPrefix uses for user-namespaced
+  // filenames: a stand-in session key so those artifacts are stored and looked up independently
+  // of whatever sessionId happened to be current when they were saved.
+  private static final String USER_NAMESPACE_SESSION_KEY = "user";
+
   private final Map<String, Map<String, Map<String, Map<String, List<Part>>>>> artifacts;
 
   public InMemoryArtifactService() {
     this.artifacts = new HashMap<>();
+  }
+
+  /**
+   * Checks if a filename uses the user namespace.
+   *
+   * @param filename Filename to check.
+   * @return true if prefixed with "user:", false otherwise.
+   */
+  private boolean fileHasUserNamespace(String filename) {
+    return filename != null && filename.startsWith("user:");
   }
 
   /**
@@ -48,7 +65,7 @@ public final class InMemoryArtifactService implements BaseArtifactService {
   public Single<Integer> saveArtifact(
       String appName, String userId, String sessionId, String filename, Part artifact) {
     List<Part> versions =
-        getArtifactsMap(appName, userId, sessionId)
+        getArtifactsMap(appName, userId, sessionId, filename)
             .computeIfAbsent(filename, unused -> new ArrayList<>());
     versions.add(artifact);
     return Single.just(versions.size() - 1);
@@ -63,7 +80,7 @@ public final class InMemoryArtifactService implements BaseArtifactService {
   public Maybe<Part> loadArtifact(
       String appName, String userId, String sessionId, String filename, @Nullable Integer version) {
     List<Part> versions =
-        getArtifactsMap(appName, userId, sessionId)
+        getArtifactsMap(appName, userId, sessionId, filename)
             .computeIfAbsent(filename, unused -> new ArrayList<>());
 
     if (versions.isEmpty()) {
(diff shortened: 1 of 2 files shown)


hemasekhar-p:
Hi @shoemoney, thank you for taking the time to submit your PR. It is currently under review by our team and we will keep you updated if any additional information is required. Thank you.

core/src/main/java/com/google/adk/artifacts/InMemoryArtifactService.java
```diff
@@ -88,10 +105,15 @@ public Maybe<Part> loadArtifact(
...
+    filenames.addAll(getArtifactsMapForSessionKey(appName, userId, sessionId).keySet());
+    // User-namespaced filenames live under the shared USER_NAMESPACE_SESSION_KEY bucket
+    // regardless of sessionId, mirroring GcsArtifactService's merged session+user listing.
+    filenames.addAll(
+        getArtifactsMapForSessionKey(appName, userId, USER_NAMESPACE_SESSION_KEY).keySet());
     return Single.just(
-        ListArtifactsResponse.builder()
-            .filenames(ImmutableList.copyOf(getArtifactsMap(appName, userId, sessionId).keySet()))
-            .build());
+        ListArtifactsResponse.builder().filenames(ImmutableList.copyOf(filenames)).build());
```
sherryfox: Python ends `list_artifact_keys` with `return sorted(filenames)`. `ImmutableList.copyOf` over a `HashSet` gives arbitrary order instead. Additionally, Java's own `GcsArtifactService` already sorts (:186), so the two backends currently disagree.

How about `ListArtifactsResponse.builder().filenames(ImmutableList.sortedCopyOf(filenames)).build());` ?

core/src/test/java/com/google/adk/artifacts/InMemoryArtifactServiceTest.java
```diff
@@ -72,6 +76,84 @@ public void saveAndReloadArtifact_reloadsArtifact() {
...
+
+    var unused =
+        service.saveArtifact(APP_NAME, USER_ID, SESSION_A, USER_FILENAME, artifact).blockingGet();
+
+    ListArtifactsResponse response =
+        service.listArtifactKeys(APP_NAME, USER_ID, SESSION_B).blockingGet();
+
+    assertThat(response.filenames()).contains(USER_FILENAME);
+  }
+
```
sherryfox, line 125: This test passes when the list would be full of ghosts so it might be worth adding one more, proving a failed load leaves nothing behind

```
@Test
public void listArtifactKeys_failedUserNamespaceLoad_doesNotCreatePhantomKey() {

Optional<Part> missing = 
  asOptional(service.loadArtifact(APP_NAME, USER_ID, SESSION_A, USER_FILENAME));

assertThat(missing).isEmpty();

ListArtifactsResponse response = 
  service.listArtifactKeys(APP_NAME, USER_ID, SESSION_B).blockingGet();

assertThat(response.filenames()).isEmpty();

}
```

and possibly similarly for listVersions?

core/src/main/java/com/google/adk/artifacts/InMemoryArtifactService.java
```diff
@@ -63,7 +80,7 @@ public Single<Integer> saveArtifact(
   public Maybe<Part> loadArtifact(
       String appName, String userId, String sessionId, String filename, @Nullable Integer version) {
     List<Part> versions =
-        getArtifactsMap(appName, userId, sessionId)
+        getArtifactsMap(appName, userId, sessionId, filename)
```
sherryfox, line 83: In a read path `computeIfAbsent` **writes an empty entry on a miss**, and `listArtifactKeys` reports `keySet()`. So a failed load **invents a file**. For reference, Python's read is pure (self.artifacts.get(path)).

This is a pre-existing issue — except this PR changes its reach. On main the ghost stays in the caller's session; now user: files share one bucket, so every session of that user sees it:

`loadArtifact(..., sessionA, "user:missing.json")  -> empty`
`listArtifactKeys(..., sessionB)                   -> ["user:missing.json"]    (main: [])`

What do you think of going back to a pure read, `ImmutableList.of()` is a shared singleton, so it allocates nothing
```
List<Part> versions = 
  getArtifactsMap(appName, userId, sessionId, filename)
    .getOrDefault(filename, ImmutableList.of());
```

core/src/main/java/com/google/adk/artifacts/InMemoryArtifactService.java
```diff
@@ -115,7 +137,7 @@ public Completable deleteArtifact(
   public Single<ImmutableList<Integer>> listVersions(
       String appName, String userId, String sessionId, String filename) {
     int size =
-        getArtifactsMap(appName, userId, sessionId)
+        getArtifactsMap(appName, userId, sessionId, filename)
             .computeIfAbsent(filename, unused -> new ArrayList<>())
             .size();
```
sherryfox, line 142: I checked and this one leaks across sessions as well. Python's `list_versions` is also a pure read. Same fix could apply:
```
int size =
  getArtifactsMap(appName, userId, sessionId, filename)
    .getOrDefault(filename, ImmutableList.of())
    .size();
```

sherryfox (changes requested):

shoemoney:
Hi @sherryfox, thanks for the review. Pushed fix covering all four points:

1. listArtifactKeys now uses ImmutableList.sortedCopyOf(filenames) to match Python sorted(filenames) and GcsArtifactService:186.
2. loadArtifact is now a pure read via getOrDefault(filename, ImmutableList.of()), no longer computeIfAbsent, so a miss does not create a phantom key visible to other sessions via the shared user: bucket.
3. listVersions same pure read fix with getOrDefault(...).size().
4. Added regression test listArtifactKeys_failedUserNamespaceLoad_doesNotCreatePhantomKey that loads a missing user: file in sessionA and asserts listArtifactKeys in sessionB stays empty.

Verified locally: InMemoryArtifactServiceTest passes 9/9 including the new test.

sherryfox (approved):

PR #1452, opened 2026-08-24
fix(chat): expose non-success HTTP status via typed exception

### LINKED ISSUE #1451: ChatCompletionsHttpClient should expose non-success HTTP status through a typed exception
## 🔴 Required Information

**Describe the Bug:**

`ChatCompletionsHttpClient` converts every non-successful HTTP response into a plain `IOException` whose message contains the OkHttp `Response.toString()` value and response body:

```java
if (!response.isSuccessful()) {
    String bodyStr = body != null ? body.string() : "";
    emitter.tryOnError(
        new IOException(
            "HTTP request failed with status: "
                + response
                + " - body: "
                + bodyStr));
}
```

As a result, callers cannot inspect the HTTP status programmatically.

For example, implementing a bounded retry for HTTP 429 currently requires parsing text such as:

```text
HTTP request failed with status: Response{protocol=h2, code=429, message=, url=...} - body: ...
```

This couples callers to:

* the exception message;
* OkHttp's `Response.toString()` format;
* field ordering inside that format;
* URL and response body content embedded in the message.

The same issue applies to both streaming and non streaming requests.

**Steps to Reproduce:**

1. Start an HTTP server whose `/v1/chat/completions` endpoint returns HTTP 429.
2. Construct a `ChatCompletionsHttpClient` pointing to that server.
3. Call `complete(...)`.
4. Inspect the emitted failure.

Example:

```java
HttpOptions options = HttpOptions.builder()
        .baseUrl("http://localhost:8080/v1")
        .build();

ChatCompletionsClient client =
        new ChatCompletionsHttpClient(options);

client.complete(request, false)
        .test()
        .assertError(IOException.class);
```

The resulting exception has no status accessor. The only representation of `429` is inside `getMessage()`.

**Expected Behavior:**

Non successful HTTP responses should be emitted through a typed exception that exposes the HTTP status separately from its presentation message.

For example:

```java
public final class ChatCompletionsHttpException extends IOException {
    private final int statusCode;
    private final String responseBody;

    public int statusCode() {
        return statusCode;
    }

    public String responseBody() {
        return responseBody;
    }
}
```

The client could then emit an exception constructed from the HTTP response, for example:

```java
new ChatCompletionsHttpException(
        response.code(),
        response.message(),
        bodyStr);
```

Alternatively, ADK could use an existing exception type that exposes a stable numeric HTTP status code.

The important requirement is that callers can inspect the HTTP status without parsing `Throwable.getMessage()`.

**Observed Behavior:**

All non success statuses are emitted as plain `IOException` instances:

```text
IOException
```

A caller cannot safely distinguish between:

* HTTP 401;
* HTTP 429;
* HTTP 500;
* a connection level `IOException`;
* an exception whose response body happens to contain `"code":429`.

The only way to determine the HTTP status is currently to parse the exception message.

**Environment Details:**

* ADK Library Version (see Maven dependency): `com.google.adk:google-adk:1.7.1`
* OS: N/A
* TS Version (`tsc --version`): N/A (Java ADK)

**Model Information:**

* Which model is being used: N/A (the issue concerns HTTP error handling in `ChatCompletionsHttpClient`, independent of the model)

---

## 🟡 Optional Information

**Regression:**

N/A. The current `ChatCompletionsHttpClient` source appears to retain the same behavior.

**Logs:**

N/A. A representative failure message is:

```text
HTTP request failed with status: Response{protocol=h2, code=429, message=, url=...} - body: ...
```

**Screenshots / Video:**

N/A.

**Additional Context:**

OpenAI compatible gateways commonly use HTTP 429 for temporary rate limiting. Applications may need to apply bounded retry/backoff specifically to that status while immediately surfacing authentication and request errors.

With a typed exception, retry code becomes stable and explicit:

```java
if (failure instanceof ChatCompletionsHttpException exception
        && exception.statusCode() == 429) {
    // Apply bounded backoff.
}
```

Without a typed exception, callers must use brittle message parsing:

```java
Pattern.compile(
    "^HTTP request failed with status: Response\\{...code=429...");
```

Message parsing can break after an OkHttp or ADK formatting change and can accidentally classify response body text as transport metadata.

### Security consideration

The current exception message includes the full response representation and response body. Depending on endpoint configuration, these may contain URL details or provider specific information that applications unintentionally write to logs.

A typed exception would allow applications to log only safe metadata, such as:

```text
statusCode=429
```

while retaining response details as explicitly accessed fields when needed.

### Scope

This request does not require ADK to implement a retry policy.

It only asks `ChatCompletionsHttpClient` to preserve machine readable HTTP response metadata so callers, or a future `OpenAiCompatibleLlm`, can implement their own retry policy safely.

```java
HttpOptions options = HttpOptions.builder()
        .baseUrl("http://localhost:8080/v1")
        .build();

ChatCompletionsClient client =
        new ChatCompletionsHttpClient(options);

client.complete(request, false)
        .test()
        .assertError(IOException.class);
```

The test server should return HTTP 429 from `/v1/chat/completions`.

**Suggested Tests:**

Add streaming and non streaming tests asserting that:

1. HTTP 429 emits the typed exception with `statusCode() == 429`.
2. HTTP 401 emits the same type with `statusCode() == 401`.
3. Connection failures remain ordinary `IOException`.
4. Response body text containing `"code":429` does not affect the exposed transport status.
5. Existing error body information remains available without being required in the exception message.

* Always (100%)

AnupamKumar-1:

- Closes: #1451

**Problem:**
`ChatCompletionsHttpClient` wraps every non-successful HTTP response (both streaming and non-streaming) in a plain `IOException` whose message embeds OkHttp's `Response#toString()` and the raw body. Callers can't inspect the HTTP status programmatically and must regex-parse `getMessage()`, which breaks if OkHttp's format changes.

**Solution:**
Added `ChatCompletionsHttpException` (extends `IOException`, so existing `catch (IOException)` code is unaffected) exposing `statusCode()` and `responseBody()`. Both the streaming and non-streaming callback paths now surface this typed exception for non-2xx HTTP responses. Connection-level failures (no HTTP response at all) still surface as plain `IOException`, since there's no status code to report.

### Testing Plan

**Unit Tests:**

Added 6 tests to `ChatCompletionsHttpClientTest`: 429/401 non-streaming typed status, streaming parity, connection failures remain plain `IOException`, response body text containing `"code":429` doesn't spoof the real status, and existing message/body content is preserved on the typed exception.

Full `core` module suite result (`./mvnw -pl core -am test`):
Tests run: 1781, Failures: 0, Errors: 0, Skipped: 24 — BUILD SUCCESS

Reproduced the original bug with a real local HTTP server (`com.sun.net.httpserver.HttpServer`) returning 429, confirmed before the fix the client only exposed `java.io.IOException` with the status buried in the message text. After the fix, the same request throws `ChatCompletionsHttpException` with `statusCode() == 429` directly accessible, no message parsing required.

### Checklist
- [x] I have read the [[CONTRIBUTING.md](https://claude.ai/chat/CONTRIBUTING.md)](./CONTRIBUTING.md) document.
- [ ] I have commented my code, particularly in hard-to-understand areas.

### Additional context
This request does not implement retry/backoff logic; it only preserves the HTTP status as machine-readable metadata so callers can implement their own retry policy against `statusCode()` instead of parsing `Response.toString()`.

### DIFF
--- a/core/src/main/java/com/google/adk/models/chat/ChatCompletionsHttpException.java
+++ b/core/src/main/java/com/google/adk/models/chat/ChatCompletionsHttpException.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.adk.models.chat;
+
+import java.io.IOException;
+
+/**
+ * Thrown by {@link ChatCompletionsHttpClient} when the server returns a non-successful HTTP status
+ * code. Extends {@link IOException} so existing {@code catch (IOException)} call sites continue to
+ * work without modification. Callers that need to branch on the HTTP status (e.g. to apply bounded
+ * retry/backoff on 429) can catch this subtype and call {@link #statusCode()} directly, without
+ * parsing {@link Throwable#getMessage()}.
+ */
+public final class ChatCompletionsHttpException extends IOException {
+  private final int statusCode;
+  private final String responseBody;
+
+  public ChatCompletionsHttpException(int statusCode, String message, String responseBody) {
+    super("HTTP request failed with status: " + message + " - body: " + responseBody);
+    this.statusCode = statusCode;
+    this.responseBody = responseBody;
+  }
+
+  public int statusCode() {
+    return statusCode;
+  }
+
+  public String responseBody() {
+    return responseBody;
+  }
+}
--- a/core/src/test/java/com/google/adk/models/chat/ChatCompletionsHttpClientTest.java
+++ b/core/src/test/java/com/google/adk/models/chat/ChatCompletionsHttpClientTest.java
@@ -36,11 +36,14 @@ import com.google.genai.types.FinishReason;
 import com.google.genai.types.FunctionCall;
 import com.google.genai.types.HttpOptions;
 import com.google.genai.types.Part;
+import io.reactivex.rxjava3.core.Flowable;
 import io.reactivex.rxjava3.subscribers.TestSubscriber;
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.time.Duration;
 import java.util.Base64;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicReference;
 import okhttp3.Call;
 import okhttp3.Callback;
 import okhttp3.MediaType;
@@ -124,6 +127,26 @@ public final class ChatCompletionsHttpClientTest {
         .model("gpt-4")
         .contents(ImmutableList.of(Content.builder().parts(Part.fromText("hello")).build()))
         .build();
+  }
+
+  private Throwable captureError(
+      Flowable<LlmResponse> flowable,
+      ArgumentCaptor<Callback> callbackCaptor,
+      Response mockResponse)
+      throws InterruptedException, IOException {
+    AtomicReference<Throwable> errorRef = new AtomicReference<>();
+    CountDownLatch latch = new CountDownLatch(1);
+    var unused =
+        flowable.subscribe(
+            resp -> {},
+            err -> {
+              errorRef.set(err);
+              latch.countDown();
+            },
+            latch::countDown);
+    callbackCaptor.getValue().onResponse(mockCall, mockResponse);
+    latch.await(AWAIT_TIMEOUT.toMillis(), MILLISECONDS);
+    return errorRef.get();
   }
 
   @Test
@@ -351,16 +374,14 @@ public final class ChatCompletionsHttpClientTest {
     ArgumentCaptor<Callback> callbackCaptor = ArgumentCaptor.forClass(Callback.class);
     doNothing().when(mockCall).enqueue(callbackCaptor.capture());
 
-    TestSubscriber<LlmResponse> testSubscriber = client.complete(minimalRequest(), false).test();
+    Throwable error =
+        captureError(client.complete(minimalRequest(), false), callbackCaptor, mockResponse);
 
-    callbackCaptor.getValue().onResponse(mockCall, mockResponse);
-    testSubscriber.await(AWAIT_TIMEOUT.toMillis(), MILLISECONDS);
-
-    testSubscriber.assertError(
-        e ->
-            e instanceof IOException
-                && e.getMessage().contains("HTTP request failed with status:")
-                && e.getMessage().contains("server exploded"));
+    assertThat(error).isInstanceOf(ChatCompletionsHttpException.class);
+    ChatCompletionsHttpException exception = (ChatCompletionsHttpException) error;
+    assertThat(exception.getMessage()).contains("HTTP request failed with status:");
+    assertThat(exception.getMessage()).contains("server exploded");
+    assertThat(exception.responseBody()).contains("server exploded");
   }
 
   /**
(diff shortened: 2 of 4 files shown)


hemasekhar-p:
Hi @AnupamKumar-1, Thanks for your effort on this PR. It is currently under review by our team and we will keep you updated if any additional information is required. Thank you.

core/src/test/java/com/google/adk/models/chat/ChatCompletionsHttpClientTest.java
```diff
@@ -393,11 +414,118 @@ public void complete_nonStreaming_propagatesEmptyBody() throws Exception {
     testSubscriber.assertError(Throwable.class);
   }
 
-  /**
-   * Verifies that caller-supplied headers reach the wire on the captured {@link Request}. This is
-   * the most common production failure mode (missing or wrong Authorization header), so it gets its
-   * own test rather than being implicit in other tests.
-   */
```
sherryfox, line 400: Was this removal intentional? 
It might be worth restoring — it's the rationale for why that test exists standalone

core/src/main/java/com/google/adk/models/chat/ChatCompletionsHttpException.java
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.adk.models.chat;
+
+import java.io.IOException;
+
```
sherryfox, line 20: Please add a javadoc

Every exception type in the repo carries a class-level javadoc, e.g. `SessionException`, `SessionNotFoundException`, `LlmCallsLimitExceededException`

core/src/test/java/com/google/adk/models/chat/ChatCompletionsHttpClientTest.java
```diff
@@ -126,6 +128,25 @@ private static LlmRequest minimalRequest() {
         .build();
   }
 
+  private Throwable captureError(
+      io.reactivex.rxjava3.core.Flowable<LlmResponse> flowable,
```
sherryfox: import Flowable rather than fully-qualifying it

core/src/test/java/com/google/adk/models/chat/ChatCompletionsHttpClientTest.java
```diff
@@ -393,11 +414,118 @@ public void complete_nonStreaming_propagatesEmptyBody() throws Exception {
...
+    Throwable error =
+        captureError(client.complete(minimalRequest(), false), callbackCaptor, mockResponse);
+
+    ChatCompletionsHttpException exception = (ChatCompletionsHttpException) error;
+    assertThat(exception.statusCode()).isEqualTo(500);
+    assertThat(exception.responseBody()).contains("\"code\":429");
+  }
+
+  @Test
+  public void complete_nonStreaming_typedException_preservesExistingMessageAndBody()
```
sherryfox: Nit: this one looks a lot like `complete_nonStreaming_propagatesHttpErrorStatus` - both use `500` / "server exploded" / contains ("HTTP request failed with status:"). The only new coverage is `responseBody()`.

Adding that one assertion to the existing test would give the same protection without a second near-identical case.

sherryfox (changes requested):
Additional request: `ChatCompletionsClient.complete()` currently documents params and return but says nothing about failure. This PR establishes a real contract — non-2xx → `ChatCompletionsHttpException`, connection failure → plain `IOException` — and the interface is where someone coding against the client would look to discover `statusCode()` exists. 

Can you please update core/src/main/java/com/google/adk/models/chat/ChatCompletionsClient.java javadoc?

AnupamKumar-1:
> Additional request: `ChatCompletionsClient.complete()` currently documents params and return but says nothing about failure. This PR establishes a real contract — non-2xx → `ChatCompletionsHttpException`, connection failure → plain `IOException` — and the interface is where someone coding against the client would look to discover `statusCode()` exists.
> 
> Can you please update core/src/main/java/com/google/adk/models/chat/ChatCompletionsClient.java javadoc?

Done, added @throws ChatCompletionsHttpException and @throws IOException to the complete() method javadoc so the failure contract is visible right at the interface level.

core/src/test/java/com/google/adk/models/chat/ChatCompletionsHttpClientTest.java
AnupamKumar-1: No, not intentional — that was an accidental removal. Restored it.

core/src/main/java/com/google/adk/models/chat/ChatCompletionsHttpException.java
AnupamKumar-1: Added a class-level javadoc following the same pattern as the other exception types in the repo.

core/src/test/java/com/google/adk/models/chat/ChatCompletionsHttpClientTest.java
AnupamKumar-1: Fixed, added the import and removed the FQN.

core/src/test/java/com/google/adk/models/chat/ChatCompletionsHttpClientTest.java
AnupamKumar-1: Agreed, the only real new coverage was responseBody(). Removed the duplicate and added that assertion to the existing complete_nonStreaming_propagatesHttpErrorStatus test instead.

sherryfox (approved):

core/src/test/java/com/google/adk/models/chat/ChatCompletionsHttpClientTest.java
```diff
@@ -126,6 +129,25 @@ private static LlmRequest minimalRequest() {
...
   }
 
+  private Throwable captureError(
+      Flowable<LlmResponse> flowable,
+      ArgumentCaptor<Callback> callbackCaptor,
+      Response mockResponse)
+      throws InterruptedException, IOException {
+    AtomicReference<Throwable> errorRef = new AtomicReference<>();
+    CountDownLatch latch = new CountDownLatch(1);
+    flowable.subscribe(
```
sherryfox: [CheckReturnValue] The result of `subscribe(...)` must be used
    flowable.subscribe(
                      ^
  If you really don't want to use the result, then assign it to a variable: `var unused = ...`.

core/src/test/java/com/google/adk/models/chat/ChatCompletionsHttpClientTest.java
```diff
@@ -393,6 +413,100 @@ public void complete_nonStreaming_propagatesEmptyBody() throws Exception {
...
+    CountDownLatch latch = new CountDownLatch(1);
+    client
+        .complete(minimalRequest(), false)
+        .subscribe(
+            resp -> {},
+            err -> {
+              errorRef.set(err);
+              latch.countDown();
+            },
+            latch::countDown);
```
sherryfox: Same as above, 
[CheckReturnValue] The result of `subscribe(...)` must be used
        .subscribe(
                  ^
  If you really don't want to use the result, then assign it to a variable: `var unused = ...`.

AnupamKumar-1:
@sherryfox Fixed the Javadoc CI failure by replacing the @throws tags with inline prose, since these exceptions aren’t declared in the method signature. No logic changes were made.

I also squashed the PR into a single commit.

Could you please retrigger CI? Thanks!

sherryfox (changes requested):
Thanks @AnupamKumar-1 for addressing my comments so eagerly 🙏

Two more, flagged by internal presubmit rather than the public CI — so it won't show up on your side. It would be great if you could update this as well

core/src/test/java/com/google/adk/models/chat/ChatCompletionsHttpClientTest.java
AnupamKumar-1: Fixed — assigned the result to var unused as suggested.

core/src/test/java/com/google/adk/models/chat/ChatCompletionsHttpClientTest.java
AnupamKumar-1: Fixed — assigned the result to var unused as suggested.

sherryfox (approved):

PR #1458, opened 2026-08-26
fix(web): make root->dev-ui redirect reverse-proxy-prefix-aware

### LINKED ISSUE #1457: AdkWebServer's root->dev-ui redirect breaks behind a path-stripping reverse proxy
## Description

`AdkWebServer.addViewControllers()` redirects `"/"` to the **root-relative** path `"/dev-ui"`:

https://github.com/google/adk-java/blob/e8b1c20d10680e7ed8e936ae1ac0ab768a0b8345/dev/src/main/java/com/google/adk/web/AdkWebServer.java#L151

```java
registry.addRedirectViewController("/", "/dev-ui");
```

When this app is deployed behind a reverse proxy that strips a path prefix before forwarding the request (a common pattern for platform-managed multi-tenant gateways, e.g. Kubernetes Gateway API `HTTPRoute` with a `URLRewrite`/`ReplacePrefixMatch` filter), the browser follows this redirect to an unprefixed path the proxy has no route for, and the dev UI 404s.

## Why this can't be fixed by the app alone with standard Spring mechanisms

Spring's own reverse-proxy support (`server.forward-headers-strategy=framework`, which installs `ForwardedHeaderFilter`) is specifically designed to solve exactly this class of problem via the `X-Forwarded-Prefix` header — but it only works for **context-relative** redirect targets (ones that do *not* start with `/`). I traced this through `ForwardedHeaderExtractingResponse#sendRedirect`:

```java
path = (path.startsWith(FOLDER_SEPARATOR) ? path :
        StringUtils.applyRelativePath(this.request.getRequestURI(), path));
```

Root-relative targets (starting with `/`, as `"/dev-ui"` does) skip the `applyRelativePath` branch entirely — the one place `X-Forwarded-Prefix` awareness would apply — so the prefix is never spliced back in, no matter how the proxy is configured.

## Proposed fix

Change the redirect target to be context-relative instead of root-relative:

```java
registry.addRedirectViewController("/", "dev-ui");
```

I verified this doesn't change behavior for the common local/non-proxied case: `RedirectView`'s `contextRelative` handling only prepends the context path when the target starts with `/` (so a context-relative target like `"dev-ui"` is passed through as-is), and the servlet container's own relative-URL resolution for `sendRedirect` still resolves `"dev-ui"` against the current request path `"/"` to `"/dev-ui"`. It *does* additionally allow `server.forward-headers-strategy=framework` + an `X-Forwarded-Prefix` header to correctly restore a proxy's path prefix, fixing the reverse-proxy case with zero app-specific workaround code.

I've opened a PR with this one-line fix plus the corresponding test update: (link to follow)

## Environment

- adk-java: current `main` (verified against commit `e8b1c20`)
- Encountered while deploying a Java ADK agent behind a Kubernetes Gateway API `HTTPRoute` that strips a per-app path prefix

luisSilva1234:
Fixes #1457

## Summary

`AdkWebServer.addViewControllers()` redirects `"/"` to the root-relative path `"/dev-ui"`. Root-relative `sendRedirect()` targets are left untouched by Spring's own reverse-proxy support (`ForwardedHeaderFilter` only applies `X-Forwarded-Prefix` to context-relative targets, via `StringUtils.applyRelativePath`, which it skips whenever the target already starts with `/`). So when this app is deployed behind a path-stripping reverse proxy, the browser follows an unprefixed `Location` header the proxy has no route for, and the dev UI 404s.

This PR makes the redirect context-relative (`"dev-ui"`) instead, so `server.forward-headers-strategy=framework` + an `X-Forwarded-Prefix` header from the proxy can correctly restore the prefix.

## Behavior for the non-proxied/local case

No change: `RedirectView`'s `contextRelative` handling only prepends the context path when the target starts with `/` (so a context-relative target is passed through as-is), and the servlet container's own relative-URL resolution for `sendRedirect` still resolves `"dev-ui"` against the current request path `"/"` to `"/dev-ui"`.

## Test plan

Added `AdkWebServerReverseProxyTest`, covering both cases with `server.forward-headers-strategy=framework` enabled:
- Without any forwarded headers: redirect is unchanged (`"dev-ui"`) — confirmed `ForwardedHeaderFilter` doesn't wrap the response at all when the request carries no `Forwarded`/`X-Forwarded-*` headers, so this matches the property being off entirely.
- With `X-Forwarded-Prefix: /my-app-prefix`: redirect becomes `http://.../my-app-prefix/dev-ui`, confirming the fix.

Also updated `AdkWebServerUITest.rootShouldRedirectToDevUi` to expect the new context-relative value (`MockHttpServletResponse.sendRedirect` stores the location verbatim with no resolution, so this assertion needed to change from `"/dev-ui"` to `"dev-ui"`).

Verified locally end-to-end (JDK 17, as required by this repo):
```
mvn -pl dev -am test
```
→ `BUILD SUCCESS`, 1776 core tests + all `dev` module tests (including the new and updated ones) pass, 0 failures.

`mvn compile` was also run to confirm `google-java-format` is happy with the new file (auto-formatted cleanly, no manual formatting issues found).

### DIFF
--- a/dev/src/main/java/com/google/adk/web/AdkWebServer.java
+++ b/dev/src/main/java/com/google/adk/web/AdkWebServer.java
@@ -143,12 +143,15 @@ public class AdkWebServer implements WebMvcConfigurer {
   }
 
   /**
-   * Configures simple automated controllers: - Redirects the root path "/" to "/dev-ui". - Forwards
-   * requests to "/dev-ui" to "/dev-ui/index.html" so the ResourceHandler serves it.
+   * Configures simple automated controllers: - Redirects the root path "/" to "dev-ui" (context-
+   * relative, not "/dev-ui", so it stays reverse-proxy-prefix-aware when this app runs behind a
+   * proxy that strips a path prefix and forwards it via server.forward-headers-strategy /
+   * X-Forwarded-Prefix). - Forwards requests to "/dev-ui" to "/dev-ui/index.html" so the
+   * ResourceHandler serves it.
    */
   @Override
   public void addViewControllers(ViewControllerRegistry registry) {
-    registry.addRedirectViewController("/", "/dev-ui");
+    registry.addRedirectViewController("/", "dev-ui");
     registry.addViewController("/dev-ui").setViewName("forward:/index.html");
     registry.addViewController("/dev-ui/").setViewName("forward:/index.html");
   }
--- a/dev/src/test/java/com/google/adk/web/AdkWebServerReverseProxyTest.java
+++ b/dev/src/test/java/com/google/adk/web/AdkWebServerReverseProxyTest.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.adk.web;
+
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.redirectedUrl;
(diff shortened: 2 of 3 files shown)


hemasekhar-p:
Hi @luisSilva1234, thank you for your contribution and We appreciate you taking the time to submit this pull request. Since this is your first contribution, could you please complete the Google CLA so that we can proceed with the review?

luissilva-mollie:
Closing in favor of #1463, opened from a single consistent GitHub account to resolve the CLA check mismatch between PR opener and commit author on this one. Same commit/content.

luisSilva1234:
Closing in favor of #1463, opened from a single consistent GitHub account to resolve the CLA check mismatch between PR opener and commit author on this one. Same commit/content.

PR #1461, opened 2026-08-27
fix: honor the user: namespace convention in InMemoryArtifactService

### LINKED ISSUE #1460: InMemoryArtifactService ignores the "user:" namespace convention — user-scoped artifacts are wrongly session-scoped
## Bug

`InMemoryArtifactService` does not implement the `user:`-prefixed filename convention that the rest of the artifact-service contract relies on: a filename starting with `user:` is meant to be stored once per `(appName, userId)` and be visible from **every** session for that user, not just the session that saved it.

Both other implementations in this repo already do this correctly:
- `GcsArtifactService.getBlobPrefix()` stores `user:`-prefixed filenames under `appName/userId/user/filename/...`, independent of `sessionId`.
- adk-python's `InMemoryArtifactService._artifact_path()` does the same (`{app_name}/{user_id}/user/{filename}`), and `list_artifact_keys` unions the session-scoped and user-namespaced prefixes.

`InMemoryArtifactService.java`, however, keys everything by `(appName, userId, sessionId, filename)` unconditionally:

```java
private Map<String, List<Part>> getArtifactsMap(String appName, String userId, String sessionId) {
  return artifacts
      .computeIfAbsent(appName, unused -> new HashMap<>())
      .computeIfAbsent(userId, unused -> new HashMap<>())
      .computeIfAbsent(sessionId, unused -> new HashMap<>());
}
```

## Repro

```java
InMemoryArtifactService service = new InMemoryArtifactService();
service.saveArtifact("app", "user1", "session-a", "user:profile.txt", artifact).blockingGet();

// Same user, different session — should see the same user-namespaced artifact.
Optional<Part> result = service.loadArtifact("app", "user1", "session-b", "user:profile.txt")
    .map(Optional::of).defaultIfEmpty(Optional.empty()).blockingGet();

// Actual: result is empty. Expected: result contains the artifact saved from session-a.
```

The same gap applies to `listArtifactKeys` (a `user:`-namespaced artifact saved elsewhere never appears) and `deleteArtifact`/`listVersions` (they silently no-op / return empty against the wrong session-scoped key instead of touching the real user-scoped entry).

## Impact

Any application relying on `InMemoryArtifactService` (the default in tests, samples, and quick local runs) for cross-session, per-user artifacts (e.g. a user profile or preference file meant to persist across conversations) silently loses that data the moment a new session starts — with no error, since the tool call still succeeds, it just operates on an empty/wrong scope.

## Fix

I have a fix + regression tests ready (keys the in-memory map the same way `GcsArtifactService` keys GCS blobs, and unions the session/user prefixes in `listArtifactKeys`) and will open a PR shortly.

Ashfaqbs:
### What is the purpose of the change

`InMemoryArtifactService` keys every artifact strictly by `(appName, userId, sessionId, filename)`, so a `user:`-prefixed filename is silently scoped to the session that saved it instead of being visible from every session for that user. Both other implementations of `BaseArtifactService` already special-case this prefix:

- `GcsArtifactService.getBlobPrefix()` stores `user:`-prefixed filenames under `appName/userId/user/filename/...`, independent of `sessionId`.
- adk-python's `InMemoryArtifactService._artifact_path()` does the same (`{app_name}/{user_id}/user/{filename}`), and its `list_artifact_keys` unions the session-scoped and user-namespaced prefixes.

Any app relying on the in-memory service (the default for tests, samples, and quick local runs) for cross-session, per-user artifacts silently loses that data the moment a new session starts, with no error — the call just succeeds against the wrong (empty) scope.

Fixes #1460.

### Brief change log

- `InMemoryArtifactService.java`: replace the 4-level nested `Map<String, Map<String, Map<String, Map<String, List<Part>>>>>` (keyed unconditionally by session) with a flat `Map<String, List<Part>>` keyed by a computed path string, mirroring `GcsArtifactService`'s blob-prefix scheme. A `user:`-prefixed filename now maps to `appName/userId/user/filename`, independent of `sessionId`; every other filename keeps the existing `appName/userId/sessionId/filename` scoping. `listArtifactKeys` now unions the session-scoped and user-namespaced prefixes, the same way `GcsArtifactService.listArtifactKeys` unions its two GCS prefixes.
- `InMemoryArtifactServiceTest.java`: three regression tests — a `user:`-namespaced artifact is visible from a different session, appears in `listArtifactKeys` for a different session, and is removed everywhere by `deleteArtifact`.

### Tests

- Added the three regression tests above; confirmed they fail against the pre-fix code (`git stash` the source change, keep the tests) and pass with the fix.
- `mvn -pl core -am test`: 1779 tests, only 1 pre-existing failure (`LocalSkillSourceTest.testListResources`, a Windows path-separator (`\` vs `/`) issue unrelated to this change — reproduces identically on `main` without this diff).

### Are these changes tested?

Yes, see above.

### Are there any user-facing changes?

Yes: `InMemoryArtifactService` now honors the `user:` filename namespace convention documented and implemented by `GcsArtifactService` and adk-python. Existing session-scoped artifacts (any filename not starting with `user:`) are unaffected.

### Was this PR generated with the help of a generative AI tool?

Yes.

Generated-by: AI coding assistant

### DIFF
--- a/core/src/main/java/com/google/adk/artifacts/InMemoryArtifactService.java
+++ b/core/src/main/java/com/google/adk/artifacts/InMemoryArtifactService.java
@@ -33,10 +33,34 @@ import org.jspecify.annotations.Nullable;
 
 /** An in-memory implementation of the {@link BaseArtifactService}. */
 public final class InMemoryArtifactService implements BaseArtifactService {
-  private final Map<String, Map<String, Map<String, Map<String, List<Part>>>>> artifacts;
+  private final Map<String, List<Part>> artifacts;
 
   public InMemoryArtifactService() {
     this.artifacts = new HashMap<>();
+  }
+
+  /**
+   * Checks if a filename uses the user namespace.
+   *
+   * @param filename Filename to check.
+   * @return true if prefixed with "user:", false otherwise.
+   */
+  private static boolean fileHasUserNamespace(String filename) {
+    return filename != null && filename.startsWith("user:");
+  }
+
+  /**
+   * Builds the storage key for an artifact.
+   *
+   * <p>A "user:"-prefixed filename is stored under a session-independent, user-scoped key so it is
+   * visible from every session for that user, matching {@link GcsArtifactService} and adk-python's
+   * {@code InMemoryArtifactService}. Any other filename is scoped to the given session, as before.
+   */
+  private static String artifactKey(
+      String appName, String userId, String sessionId, String filename) {
+    return fileHasUserNamespace(filename)
+        ? String.format("%s/%s/user/%s", appName, userId, filename)
+        : String.format("%s/%s/%s/%s", appName, userId, sessionId, filename);
   }
 
   /**
@@ -48,8 +72,8 @@ public final class InMemoryArtifactService implements BaseArtifactService {
   public Single<Integer> saveArtifact(
       String appName, String userId, String sessionId, String filename, Part artifact) {
     List<Part> versions =
-        getArtifactsMap(appName, userId, sessionId)
-            .computeIfAbsent(filename, unused -> new ArrayList<>());
+        artifacts.computeIfAbsent(
+            artifactKey(appName, userId, sessionId, filename), unused -> new ArrayList<>());
     versions.add(artifact);
     return Single.just(versions.size() - 1);
   }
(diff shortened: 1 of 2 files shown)


hemasekhar-p:
Hi @Ashfaqbs, thank you for your contribution! We appreciate you taking the time to submit this pull request. Currently this PR is under review by our team, we will keep you posted if any additional information is required. thank you.

sherryfox:
Thanks so much for putting this PR together, @Ashfaqbs! We really appreciate your time and effort in diagnosing and fixing this.

As mentioned in #1460, a concurrent fix for this bug was recently merged via #1449. While your approach here is a completely valid and defensible port (mirroring `adk-python`), the solution in #1449 was already in flight and has now landed.

Please don't let this discourage you — your diagnosis was spot on, and we would love to welcome more contributions from you in the future!

Issue #1280, opened 2026-06-16
[BUG] `ConversionUtils.adkToMcpToolType` throws `NoSuchElementException` for a tool declaration without parameters


SuperCorleone:
## Describe the Bug
`adkToMcpToolType` guards the empty-*declaration* case but then calls `toolDeclaration.get().parameters().get()` — an unguarded `Optional.get()` on `parameters()`. A declaration that is present but has no parameters (a valid no-argument tool) yields `NoSuchElementException`. All built-in ADK tools populate `parameters`, so the library itself never triggers it, but a user-supplied custom tool overriding `declaration()` without parameters would.

## Root Cause
```java
// ConversionUtils.adkToMcpToolType, ~L36
... toolDeclaration.get().parameters().get() ...   // NoSuchElementException when parameters is empty
```

## Suggested Fix
Guard parameters: `toolDeclaration.get().parameters().orElse(null)` (or build the MCP tool without an input schema when parameters are absent — which is exactly what the failing test expects).

## Corresponding Test (generated)
```java
@Test
public void testAdkToMcpToolType_WithNonEmptyDeclarationAndNullSchema_ReturnsToolWithoutInputSchema() {
    // Given
    String toolName = "testTool";
    String toolDescription = "A test tool";

    when(mockBaseTool.name()).thenReturn(toolName);
    when(mockBaseTool.description()).thenReturn(toolDescription);
    when(mockBaseTool.declaration()).thenReturn(Optional.of(mockFunctionDeclaration));
    when(mockFunctionDeclaration.parameters()).thenReturn(Optional.empty());

    // When
    McpSchema.Tool result = conversionUtils.adkToMcpToolType(mockBaseTool);

    // Then
    assertNotNull(result);
    assertEquals(toolName, result.name());
    assertEquals(toolDescription, result.description());
    assertNull(result.inputSchema());
}
```

## Environment
- Project: adk / google-adk (`com.google.adk`) · upstream: github.com/google/adk-java
- Version: `0.5.1-SNAPSHOT` (local) · Commit: `<fill SUT commit>` · Java 17 · OS macOS

This input was generated by the test case generator `TestFusion` developed in our [STAR](https://star.inf.usi.ch/#/home) lab.

hemasekhar-p:
Hi @SuperCorleone, Thank you for reporting this issue. I am able to reproduce this issue which you described. Currently this Issue is under review by our team, we will keep you update if any additional information is required. thank you.

[...]

Issue #1279, opened 2026-06-16
[BUG] `GoogleSearchTool.processLlmRequest` throws `NoSuchElementException` synchronously on an empty model


SuperCorleone:
## Describe the Bug
`processLlmRequest` does `String model = llmRequestBuilder.build().model().get();` — an unguarded `Optional.get()`. If the model is absent it throws `NoSuchElementException` **synchronously**, escaping the returned `Completable` instead of being delivered via `onError`. The method's own `else` branch is written to handle a null model gracefully (returns `Completable.error(IllegalArgumentException)`), and the sibling `GoogleMapsTool.java:67` uses the correct `model().orElse(null)` — so this is an inconsistency/fragility. In the normal flow the `Basic` request processor sets the model before any tool runs, so it isn't reached there.

## Root Cause
```java
String model = llmRequestBuilder.build().model().get();   // NoSuchElementException if empty; thrown synchronously
```

## Suggested Fix
Use `model().orElse(null)` (matching `GoogleMapsTool`) so the existing null-handling branch runs and errors flow through the `Completable`.

## Corresponding Test (generated)
```java
@Test
public void testProcessLlmRequest_Gemini2ModelWithNoModel_ThrowsException() {
    // Given
    GoogleSearchTool tool = GoogleSearchTool.INSTANCE;
    LlmRequest.Builder requestBuilder = LlmRequest.builder();

    // When & Then
    Completable result = tool.processLlmRequest(requestBuilder, null);
    
    // Use a proper observer to capture the error
    AtomicReference<Throwable> capturedException = new AtomicReference<>();
    result.subscribe(
        new CompletableObserver() {
          @Override
          public void onSubscribe(Disposable d) {}
          
          @Override
          public void onComplete() {}
          
          @Override
          public void onError(Throwable e) {
            capturedException.set(e);
          }
        }
    );
    
    assertNotNull(capturedException.get(), "Expected an exception to be thrown");
    assertTrue(capturedException.get() instanceof IllegalArgumentException);
    assertTrue(capturedException.get().getMessage().contains("Google search tool is not supported"));
}
```

## Environment
- Project: adk / google-adk (`com.google.adk`) · upstream: github.com/google/adk-java
- Version: `0.5.1-SNAPSHOT` (local) · Commit: `<fill SUT commit>` · Java 17 · OS macOS

This input was generated by the test case generator `TestFusion` developed in our [STAR](https://star.inf.usi.ch/#/home) lab.

[...]

PR #1465, opened 2026-08-27
fix(memory): guard parts without text in searchMemory

### LINKED ISSUE #1464: [BUG] InMemoryMemoryService.searchMemory throws NoSuchElementException for events with non-text parts
## 🔴 Required Information

**Describe the Bug:**

`InMemoryMemoryService.searchMemory` calls `part.text().get()` without checking whether the `Optional` holds a value:

```java
// InMemoryMemoryService.java:108
for (Part part : event.content().get().parts().get()) {
  if (!Strings.isNullOrEmpty(part.text().get())) {   // NoSuchElementException for a non-text part
    Matcher matcher = WORD_PATTERN.matcher(part.text().get());
```

`Part.text()` is empty for `functionCall`, `functionResponse`, `inlineData` and `fileData` parts. `addSessionToMemory` stores those events — it only drops events whose `parts` list is empty. `searchMemory` then scans every session stored under the same app and user, so one tool call in any one session makes every later search for that app and user throw, whatever the query.

**Steps to Reproduce:**

1. Store a session holding an event with a `functionCall` part. No model or network is involved.
2. Search that app and user.

```java
InMemoryMemoryService memoryService = new InMemoryMemoryService();

Session session =
    Session.builder("session")
        .appName("app")
        .userId("user")
        .events(
            ImmutableList.of(
                Event.builder()
                    .author("agent")
                    .content(
                        Content.fromParts(
                            Part.builder()
                                .functionCall(FunctionCall.builder().name("get_weather").build())
                                .build()))
                    .build()))
        .build();

memoryService.addSessionToMemory(session).blockingAwait();
memoryService.searchMemory("app", "user", "weather").blockingGet();
```

**Expected Behavior:**

Parts without text are skipped and the search returns the entries whose text matched.

**Observed Behavior:**

```text
java.util.NoSuchElementException: No value present
	at java.base/java.util.Optional.get(Optional.java:143)
	at com.google.adk.memory.InMemoryMemoryService.lambda$searchMemory$5(InMemoryMemoryService.java:108)
	at io.reactivex.rxjava3.internal.operators.single.SingleFromCallable.subscribeActual(SingleFromCallable.java:43)
	at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4855)
	at io.reactivex.rxjava3.core.Single.blockingGet(Single.java:3644)
```

**Environment Details:**

- ADK Library Version: 1.8.1-SNAPSHOT (`main`), also present in 1.8.0
- OS: macOS 26.4.1
- TS Version: N/A

**Model Information:**

- N/A

---

## 🟡 Optional Information

**Regression:**

No. The unguarded `.get()` has been there since the class was added in `c8fed2d` (2025-06-28).

**Additional Context:**

Reachable from an agent: `LoadMemoryTool` → `ToolContext.searchMemory` → `BaseMemoryService.searchMemory`. The trigger is an app that saves a tool-using session with `addSessionToMemory` and later lets the agent call `load_memory`. `InMemoryMemoryService` is the default in `InMemoryRunner` and in the dev server, so this is the path most people are on.

Every other `part.text()` read in `core` guards with `isPresent()`, `filter` or `orElse` (e.g. `LlmAgent`, `Claude`, `LoggingPlugin`, `CodeExecutionUtils`; `LlmEventSummarizer` uses `part.text().filter(not(String::isEmpty))` for the same job). This is the one that does not.

adk-python filters the same loop on `if part.text`, and adk-go skips with `if part.Text == ""`.

Same class of bug as #1279 and #1280.

- Always (100%) once a stored session holds a part without text.

I have a fix and tests ready and will open a PR.

1wos:

- Closes: #1464

**Problem:**

`InMemoryMemoryService.searchMemory` calls `part.text().get()` before checking that the `Optional` holds a value. `Part.text()` is empty for `functionCall`, `functionResponse`, `inlineData` and `fileData` parts, so the call throws `NoSuchElementException` and the whole search fails.

`addSessionToMemory` stores those events — it only drops events whose `parts` list is empty. `searchMemory` scans every session stored under the same app and user, so one tool call in any one session breaks every later search for that app and user, whatever the query. Reachable from an agent through `LoadMemoryTool` → `ToolContext.searchMemory`.

**Solution:**

Read the text through `orElse("")` and keep the existing empty-text check, so a part without text is skipped exactly like a part whose text is blank.

```java
String text = part.text().orElse("");
if (!text.isEmpty()) {
  Matcher matcher = WORD_PATTERN.matcher(text);
```

Every other `part.text()` read in `core` already guards with `isPresent()`, `filter` or `orElse` — `LlmEventSummarizer` uses `part.text().filter(not(String::isEmpty))` for the same job. This was the one that did not. It also matches adk-python, which filters the same loop on `if part.text`, and adk-go, which skips with `if part.Text == ""`.

### Testing Plan

**Unit Tests:**

`InMemoryMemoryService` had no test class. Added five cases, one per part type without text plus a mixed event:

- `searchMemory_functionCallPart_skipsPartInsteadOfThrowing`
- `searchMemory_functionResponsePart_skipsPartInsteadOfThrowing`
- `searchMemory_inlineDataPart_skipsPartInsteadOfThrowing`
- `searchMemory_fileDataPart_skipsPartInsteadOfThrowing`
- `searchMemory_textAndFunctionCallInOneEvent_stillMatchesOnTheText`

All five fail on `main` with the new test class alone:

```
$ ./mvnw -pl core test -Dtest=InMemoryMemoryServiceTest -DfailIfNoTests=false
[ERROR] Tests run: 5, Failures: 0, Errors: 5, Skipped: 0
java.util.NoSuchElementException: No value present
	at java.base/java.util.Optional.get(Optional.java:143)
	at com.google.adk.memory.InMemoryMemoryService.lambda$searchMemory$5(InMemoryMemoryService.java:108)
	at io.reactivex.rxjava3.core.Single.blockingGet(Single.java:3644)
```

With the fix:

```
$ ./mvnw -pl core test -Dtest=InMemoryMemoryServiceTest -DfailIfNoTests=false
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Full `core` suite on this branch, rebased onto `main`:

```
$ ./mvnw -f core/pom.xml test
[WARNING] Tests run: 1794, Failures: 0, Errors: 0, Skipped: 24
[INFO] BUILD SUCCESS
```

Ran a driver locally (not part of this PR) modelled on `AgentWithMemoryTest`, but with the agent calling a tool in the first turn: `InMemoryRunner` + `LlmAgent` with a `getWeather` `FunctionTool` and `LoadMemoryTool`, driven by `TestLlm` so the run is deterministic and needs no model credentials. Turn one asks about the weather and the agent uses the tool; the session is saved with `addSessionToMemory`; turn two makes the agent call `load_memory`.

`AgentWithMemoryTest` passes today only because its first turn is text-only. Adding a tool call to that first turn is enough to break it.

On `main`:

```
turn 1 stored 4 events:
  user           text
  weather_agent  functionCall
  weather_agent  functionResponse
  weather_agent  text

[ERROR] MemoryE2eDriverTest.agentUsesToolThenSearchesMemory » NoSuchElement No value present
```

With the fix:

```
turn 1 stored 4 events:
  user           text
  weather_agent  functionCall
  weather_agent  functionResponse
  weather_agent  text

load_memory returned: FunctionResponse{name=loadMemory,
  response={memories=[{content={parts=[{text=what is the weather in Seoul}],
  role=user}, author=user}]}}
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
```

Any app that saves a tool-using session with `addSessionToMemory` and later lets the agent call `load_memory` hits this today.

This fixes the crash only. `InMemoryMemoryService` has drifted from adk-python in other ways since — result ranking, Unicode-aware word extraction, thread safety — and none of that is in scope here.

### Checklist

### Additional context

Same class of bug as #1279 and #1280. Follows the "Alignment with adk-python" section of CONTRIBUTING.md.

### DIFF
--- a/core/src/test/java/com/google/adk/memory/InMemoryMemoryServiceTest.java
+++ b/core/src/test/java/com/google/adk/memory/InMemoryMemoryServiceTest.java
@@ -1,0 +1,140 @@
+/*
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.adk.memory;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import com.google.adk.events.Event;
+import com.google.adk.sessions.Session;
+import com.google.common.collect.ImmutableList;
+import com.google.genai.types.Content;
+import com.google.genai.types.FileData;
+import com.google.genai.types.FunctionCall;
+import com.google.genai.types.FunctionResponse;
+import com.google.genai.types.Part;
+import java.util.Map;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Tests for {@link InMemoryMemoryService}. */
+@RunWith(JUnit4.class)
+public final class InMemoryMemoryServiceTest {
+
+  private static final String APP_NAME = "app";
+  private static final String USER_ID = "user";
+
+  private static Event event(Part... parts) {
+    return Event.builder().author("agent").content(Content.fromParts(parts)).build();
+  }
+
+  private static Session sessionWith(Event... events) {
+    return Session.builder("session")
+        .appName(APP_NAME)
+        .userId(USER_ID)
+        .events(ImmutableList.copyOf(events))
+        .build();
+  }
+
+  @Test
+  public void searchMemory_functionCallPart_skipsPartInsteadOfThrowing() {
+    Part functionCall =
+        Part.builder().functionCall(FunctionCall.builder().name("loadMemory").build()).build();
+    InMemoryMemoryService service = new InMemoryMemoryService();
+    service
+        .addSessionToMemory(
+            sessionWith(event(Part.fromText("weather in Seoul")), event(functionCall)))
+        .blockingAwait();
+
+    SearchMemoryResponse response =
+        service.searchMemory(APP_NAME, USER_ID, "weather").blockingGet();
+
+    assertThat(response.memories()).hasSize(1);
+  }
+
+  @Test
+  public void searchMemory_functionResponsePart_skipsPartInsteadOfThrowing() {
+    Part functionResponse =
+        Part.builder()
+            .functionResponse(
+                FunctionResponse.builder()
+                    .name("loadMemory")
+                    .response(Map.of("result", "ok"))
+                    .build())
+            .build();
+    InMemoryMemoryService service = new InMemoryMemoryService();
+    service.addSessionToMemory(sessionWith(event(functionResponse))).blockingAwait();
+
+    SearchMemoryResponse response =
+        service.searchMemory(APP_NAME, USER_ID, "weather").blockingGet();
+
+    assertThat(response.memories()).isEmpty();
+  }
+
+  @Test
+  public void searchMemory_inlineDataPart_skipsPartInsteadOfThrowing() {
+    Part inlineData = Part.fromBytes(new byte[] {1, 2, 3}, "image/png");
+    InMemoryMemoryService service = new InMemoryMemoryService();
+    service
+        .addSessionToMemory(
+            sessionWith(event(Part.fromText("weather in Seoul")), event(inlineData)))
+        .blockingAwait();
+
+    SearchMemoryResponse response =
+        service.searchMemory(APP_NAME, USER_ID, "weather").blockingGet();
+
+    assertThat(response.memories()).hasSize(1);
+  }
+
+  @Test
+  public void searchMemory_fileDataPart_skipsPartInsteadOfThrowing() {
+    Part fileData =
+        Part.builder()
+            .fileData(
+                FileData.builder()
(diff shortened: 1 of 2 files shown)


hemasekhar-p:
Hi @1wos, thank you for your contribution! We appreciate you taking the time to submit this pull request. Currently this PR is under review by our team we will keep you posted if any additional information is required. thank you.

core/src/test/java/com/google/adk/memory/InMemoryMemoryServiceTest.java
```diff
@@ -0,0 +1,103 @@
...
+import com.google.genai.types.FunctionResponse;
+import com.google.genai.types.Part;
+import java.util.Map;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Tests for {@link InMemoryMemoryService}. */
+@RunWith(JUnit4.class)
+public final class InMemoryMemoryServiceTest {
```
sherryfox, line 36: Both the issue and the PR body list four part types with no text — `functionCall`, `functionResponse`, `inlineData`, `fileData`. These cover the first two.

I checked the other two on unpatched main and they hit the same `NoSuchElementException`, and your fix handles them, so this is only a coverage gap — but it's the kind that would let a later refactor narrow the guard to function parts without anything failing. Could you please cover them?

sherryfox (changes requested):

core/src/test/java/com/google/adk/memory/InMemoryMemoryServiceTest.java
1wos: Added both. They hit the same NoSuchElementException on unpatched main, so all four part types from the issue are covered now and a narrowed guard would fail the suite. Full core run is at 1794 passing.

sherryfox (approved):

PR #1467, opened 2026-08-28
fix(sessions): validate HTTP status before parsing Vertex AI session responses

### LINKED ISSUE #1466: VertexAiClient ignores HTTP status, so session GET errors become phantom empty Sessions
## 🔴 Required Information

**Describe the Bug:**

`VertexAiClient.getJsonResponse` never inspects the HTTP status of the API response. Any non-empty response body is parsed and returned as a success, and an empty body returns `Maybe.empty()`. As a result `VertexAiSessionService.getSession` cannot distinguish three different outcomes:

1. the session genuinely does not exist (404),
2. the request failed for another reason (403 / 429 / 500 with a JSON error body),
3. the request failed with an empty body.

In cases 1 and 2 it constructs a `Session` from whatever fields it can find in the body — requested id, empty state, and `lastUpdateTime` null via `.orElse(null)`. Callers observe a plausible-looking empty session instead of a miss or an error.

This is dangerous for any get-or-create pattern. A transient 500 is indistinguishable from "session not found", so an application that auto-creates on a miss will silently create a *new* session mid-conversation and fork the user's history, discarding prior turns. There is no signal at the call site that anything went wrong.

Both locations on `main`:

- `core/src/main/java/com/google/adk/sessions/VertexAiClient.java#L208-L219` — line 216 returns `Maybe.empty()` for an empty body; line 219 parses any non-empty body regardless of status.
- `core/src/main/java/com/google/adk/sessions/VertexAiSessionService.java#L197-L221` — lines 220-221 read `updateTime` via `.orElse(null)`, producing the null-timestamp shape.

**Steps to Reproduce:**

1. Configure `VertexAiSessionService` against a Vertex AI reasoning engine.
2. Call `getSession(appName, userId, sessionId, Optional.empty())` for a session id that does not exist. Observe a `Session` is returned rather than an empty `Maybe`.
3. Cause the GET to fail with a non-404 status that returns a JSON error body — e.g. revoke the caller's permission on the reasoning engine (403), or exceed quota (429). Observe a `Session` is returned again, with the same shape.
4. Note that steps 2 and 3 are indistinguishable from each other, and from a real session that happens to have no state.

**Expected Behavior:**

`getSession` should reflect the transport outcome:

- 404 / NOT_FOUND → empty `Maybe` (a real miss)
- any other non-2xx → an error signalled to the caller
- 2xx → the parsed `Session`

**Observed Behavior:**

All three produce a `Session` with the requested id, empty state, and `lastUpdateTime == null`. No exception, no log, nothing distinguishing them.

**Environment Details:**

- ADK Library Version: reproduced on `1.2.0`; code inspected on `main` after `1.8.0` and unchanged
- OS: macOS and Linux
- TS Version: N/A (Java)

**Model Information:**

- N/A — this is in the session service transport layer, independent of model choice.

---

## 🟡 Optional Information

**Regression:**

No — present in `1.2.0` and still present on `main` today. Note `getSession` did gain a backend-owner check since `1.2.0` ("deny as not-found so existence is not revealed"), but the status handling was not addressed.

**Additional Context:**

We hit this in production. Across a month of traffic our "session not found" path never fired once, because every miss was being converted into a phantom session. Separately, a burst of `VertexAiClient` NPEs correlated day-for-day with a spike in requests arriving with an empty session id, as callers fell back after receiving degraded responses.

Workarounds based on the returned object's *shape* are not reliable, which is why we think this needs fixing in the library:

- checking for a `USER_ID` entry in state fails, because legitimately state-less sessions exist (we use short-lived sessions for content moderation);
- checking for a null `lastUpdateTime` alone fails, because 403 / 429 / 500 produce the same shape as a 404.

We currently wrap the service with a decorator that re-probes with a status-checking client and maps NOT_FOUND to an empty `Maybe`, propagating every other status as an error. That works but duplicates the transport layer.

Two adjacent cases we did **not** attempt to work around, in case they are the same root cause: the ADK collapses event-list errors into an empty conversation history, and `appendEvent` failures are swallowed with a warn log.

Always (100%) — it is deterministic for any non-2xx response on the session GET path.

AnupamKumar-1:

- Closes: #1466

**Problem:**
`VertexAiClient.getJsonResponse` never checks the HTTP status of the Vertex AI Session API response. It only checks whether the body is empty. So a non-2xx response with a body (403, 429, 500, etc.) gets parsed exactly like a successful one, and `VertexAiSessionService.getSession` builds a phantom `Session` — with the requested id, empty state, and a null `lastUpdateTime` (serialized as `0.0` by `Session.toJson()`) — instead of surfacing the failure. This is indistinguishable from a real session that happens to have no events, and callers using a get-or-create pattern can silently start a new session mid-conversation on a transient error.

**Solution:**
Added `getStatusCode()` to `ApiResponse`/`HttpApiResponse`, and updated `getJsonResponse` to branch on status before touching the body:
- `404` → treated as a genuine miss, returns empty.
- Any other non-2xx → returns a new `VertexAiApiException(statusCode, body)` instead of parsing.
- `2xx` with an empty body → empty (unchanged).
- `2xx` with a body → parsed and returned (unchanged).

### Testing Plan

**Unit Tests:**

Summary: `VertexAiSessionServiceTest` — 32 tests, 0 failures. Added `getSession_permissionDenied_propagatesAsError` (403) and `getSession_serverError_propagatesAsError` (500), both asserting a `VertexAiApiException` is thrown instead of a phantom session. Renamed `getEmptySession_success` to `getEmptySession_returnsNull` and updated `getAndDeleteSession_success` to assert a real 404 now correctly returns `null` from `getSession()` rather than throwing a mock-only exception. Also ran the full module test suite (`mvn -pl core -am test`): 1791 tests, 0 failures.

Not applicable — this is a transport-layer fix inside `VertexAiSessionService`/`VertexAiClient` with no UI surface, exercised entirely through the unit tests above.

### Checklist

### Additional context

None.

### DIFF
--- a/core/src/main/java/com/google/adk/sessions/HttpApiResponse.java
+++ b/core/src/main/java/com/google/adk/sessions/HttpApiResponse.java
@@ -35,6 +35,11 @@ public final class HttpApiResponse extends ApiResponse {
     return response.body();
   }
 
+  @Override
+  public int getStatusCode() {
+    return response.code();
+  }
+
   /** Closes the Http response. */
   @Override
   public void close() {
--- a/core/src/main/java/com/google/adk/sessions/VertexAiApiException.java
+++ b/core/src/main/java/com/google/adk/sessions/VertexAiApiException.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.adk.sessions;
+
+/**
+ * Signals a non-2xx, non-404 HTTP status from the Vertex AI Session API. Extends {@link
+ * SessionException} so existing {@code catch (SessionException)} call sites still work.
+ */
+public final class VertexAiApiException extends SessionException {
+  private final int statusCode;
+
+  VertexAiApiException(int statusCode, String responseBody) {
+    super(
+        "Vertex AI Session API request failed with HTTP status "
+            + statusCode
+            + (responseBody == null || responseBody.isEmpty() ? "" : ": " + responseBody));
+    this.statusCode = statusCode;
+  }
+
+  /** Returns the HTTP status code returned by the API. */
+  public int statusCode() {
+    return statusCode;
+  }
+}
--- a/core/src/main/java/com/google/adk/sessions/VertexAiClient.java
+++ b/core/src/main/java/com/google/adk/sessions/VertexAiClient.java
@@ -206,16 +206,29 @@ final class VertexAiClient {
    */
   @Nullable
   private static Maybe<JsonNode> getJsonResponse(ApiResponse apiResponse) {
+    if (apiResponse == null) {
+      return Maybe.empty();
+    }
     try {
-      if (apiResponse == null || apiResponse.getResponseBody() == null) {
+      int statusCode = apiResponse.getStatusCode();
+      String responseString;
+      try {
+        ResponseBody responseBody = apiResponse.getResponseBody();
+        responseString = responseBody == null ? "" : responseBody.string();
+      } catch (IOException e) {
+        return Maybe.error(new UncheckedIOException(e));
+      }
+
+      if (statusCode == 404) {
+        return Maybe.empty();
+      }
+      if (statusCode < 200 || statusCode >= 300) {
+        return Maybe.error(new VertexAiApiException(statusCode, responseString));
+      }
+      if (responseString.isEmpty()) {
         return Maybe.empty();
       }
       try {
-        ResponseBody responseBody = apiResponse.getResponseBody();
-        String responseString = responseBody.string(); // Read body here
-        if (responseString.isEmpty()) {
-          return Maybe.empty();
-        }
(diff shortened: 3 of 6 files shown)


hemasekhar-p:
Hi @AnupamKumar-1,  Thank you for you contribution and We appreciate you taking the time to submit this PR. through local testing that your fix addresses issue. Currently, the PR is under review by our team and we will keep you updated if any additional information is required. Thank you.

sherryfox (changes requested):
Thanks for the fix @AnupamKumar-1!

One mechanical thing: squash the commits since 2 on a PR will fail our CI atm.

AnupamKumar-1:
> Thanks for the fix @AnupamKumar-1!
> 
> One mechanical thing: squash the commits since 2 on a PR will fail our CI atm.

@sherryfox Done, squashed to one commit.

sherryfox (approved):

core/src/main/java/com/google/adk/sessions/ApiResponse.java
```diff
@@ -23,6 +23,9 @@ public abstract class ApiResponse implements AutoCloseable {
   /** Gets the HttpEntity. */
   public abstract ResponseBody getResponseBody();
 
+  /** Gets the HTTP status code of the response. */
+  public abstract int getStatusCode();
```
damianmomotgoogle: This class is public. To maintain backward compatibility (in case some other project is extending it) we should introduce it with a default implementation - might be throwing unimplemented error

otherwise potential existing sub-classes will stop compiling after version bump

damianmomotgoogle (changes requested):

core/src/main/java/com/google/adk/sessions/ApiResponse.java
AnupamKumar-1: Addressed — `getStatusCode()` is now a concrete method with a default `UnsupportedOperationException` implementation instead of being `abstract`, so existing subclasses won't break on upgrade.

Added a regression test (`ApiResponseTest`) to lock in this behavior.

damianmomotgoogle (approved):

PR #1468, opened 2026-08-28
fix(memory): format search timestamps as epoch millis

rootkiller6788:
## Summary

`InMemoryMemoryService.formatTimestamp` used `Instant.ofEpochSecond`, but `Event.timestamp()` is epoch milliseconds (`Event.build()` defaults to `Instant.now().toEpochMilli()`), so returned memory timestamps were ~1000x in the future.

## Changes

- Format memory timestamps from epoch millis via `Instant.ofEpochMilli`.
- Add `InMemoryMemoryServiceTimestampTest` covering the regression.

## Verification

- New tests passes: `InMemoryMemoryServiceTimestampTest` (1 test).
- `AgentWithMemoryTest` still passes.

Note: the non-text-part crash fix originally in this PR was dropped — it landed separately in #1465.

### DIFF
--- a/core/src/main/java/com/google/adk/memory/InMemoryMemoryService.java
+++ b/core/src/main/java/com/google/adk/memory/InMemoryMemoryService.java
@@ -136,6 +136,6 @@ public final class InMemoryMemoryService implements BaseMemoryService {
   }
 
   private String formatTimestamp(long timestamp) {
-    return Instant.ofEpochSecond(timestamp).toString();
+    return Instant.ofEpochMilli(timestamp).toString();
   }
 }
--- a/core/src/test/java/com/google/adk/memory/InMemoryMemoryServiceTimestampTest.java
+++ b/core/src/test/java/com/google/adk/memory/InMemoryMemoryServiceTimestampTest.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


sherryfox:
Hi @rootkiller6788, thank you for your contribution and for taking the time to submit this PR. 

You're right that `InMemoryMemoryService.searchMemory` has two bugs. The first fix unfortunately duplicates #1465, which should land soon. The second is definitely needed.

One ask for the scope then: once #1465 lands, could you rebase this down to just the `formatTimestamp` change? That drops the conflict and leaves a small, focused change that's easy to land — right now both PRs add the same new test file.

rootkiller6788:
Thanks for the review, @sherryfox. I've narrowed the PR down to just the formatTimestamp fix as you suggested.

- Dropped the non-text part guard from this branch — it's covered by #1465, so keeping it here would duplicate the fix.
- Removed `InMemoryMemoryServiceTest.java` from this PR. Since #1465 also adds a file under that name, I moved the timestamp regression test into its own `InMemoryMemoryServiceTimestampTest.java` so there's no conflict on the test file once #1465 lands.
- The production change is now a single line: `formatTimestamp` reads the event timestamp as epoch milliseconds via `Instant.ofEpochMilli`, matching `Event.timestamp()`.

The timestamp test passes against the core module (1 test, no failures). Could you please take another look?

sherryfox:
Two mechanical things @rootkiller6788: `check-commit-count` is red at 2 commits, and the title still says "skip non-text parts" — worth retitling, since the title becomes the commit message and release-please turns it into the changelog entry.

rootkiller6788:
Fix: memory search timestamps were parsed as seconds, not millis

  InMemoryMemoryService.formatTimestamp() was decoding timestamps with Instant.ofEpochSecond(...). The timestamps going
  through it are epoch millis, so the formatter read them as seconds and search result times came out wrong — anything
  recent showed up as a date in 1970.

  Changed it to Instant.ofEpochMilli(...), which is what the values actually are.

  Tests
  - Added InMemoryMemoryServiceTimestampTest covering the timestamp formatting path.

  This is the narrowed version of the earlier fix — just the timestamp bug, nothing else.

sherryfox (approved):

PR #1485, opened 2026-09-08
fix(runner): default InMemoryRunner appName to match Python SDK

### LINKED ISSUE #1486: InMemoryRunner default appName should be "InMemoryRunner" to match adk-python
## Description

`InMemoryRunner(BaseAgent)` currently defaults `appName` to `agent.name()`.

In [adk-python](https://github.com/google/adk-python/blob/main/src/google/adk/runners.py), the default is `"InMemoryRunner"` when no `app_name` is provided.

There is an explicit TODO in Java:
https://github.com/google/adk-java/blob/main/core/src/main/java/com/google/adk/runner/InMemoryRunner.java

## Expected behavior

`new InMemoryRunner(agent)` should use `appName = "InMemoryRunner"`.

## Actual behavior

`appName` defaults to `agent.name()`.

## Impact

Callers using `runAsync(userId, sessionId, ...)` may fail to find sessions created under a different app name than the runner's implicit default.

## Proposed fix

PR #1485

mithun-sudo:
Fixes #1486 

## Summary
- Default `InMemoryRunner(BaseAgent)` appName to `"InMemoryRunner"` instead of `agent.name()`, matching [adk-python](https://github.com/google/adk-python/blob/main/src/google/adk/runners.py).
- Add `InMemoryRunnerTest` and update `AgentWithMemoryTest` to use `runner.appName()`.

## Breaking change (minor)
Callers that relied on implicit `agent.name()` as appName should use `new InMemoryRunner(agent, agent.name())` explicitly.

## Test plan
- [x] `./mvnw -pl core -am test` — BUILD SUCCESS

### DIFF
--- a/contrib/spring-ai/src/test/java/com/google/adk/models/springai/SpringAIIntegrationTest.java
+++ b/contrib/spring-ai/src/test/java/com/google/adk/models/springai/SpringAIIntegrationTest.java
@@ -75,7 +75,7 @@ class SpringAIIntegrationTest {
     // when
     Runner runner = new InMemoryRunner(agent);
     Session session =
-        runner.sessionService().createSession(agent.name(), "test-user").blockingGet();
+        runner.sessionService().createSession(runner.appName(), "test-user").blockingGet();
 
     Content userMessage =
         Content.builder().role("user").parts(List.of(Part.fromText("What is a qubit?"))).build();
@@ -152,7 +152,7 @@ class SpringAIIntegrationTest {
     // when
     Runner runner = new InMemoryRunner(agent);
     Session session =
-        runner.sessionService().createSession(agent.name(), "test-user").blockingGet();
+        runner.sessionService().createSession(runner.appName(), "test-user").blockingGet();
 
     Content userMessage =
         Content.builder()
@@ -221,7 +221,7 @@ class SpringAIIntegrationTest {
     // when
     Runner runner = new InMemoryRunner(agent);
     Session session =
-        runner.sessionService().createSession(agent.name(), "test-user").blockingGet();
+        runner.sessionService().createSession(runner.appName(), "test-user").blockingGet();
 
     Content userMessage =
         Content.builder()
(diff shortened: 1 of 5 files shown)


hemasekhar-p:
Hi @mithun-sudo, thank you for your contribution and  We appreciate you taking the time to submit this pull request. I have noticed that the maven build is failing with your changes, could you please address this?

mithun-sudo:
Fixed CI: spring-ai tests were still creating sessions with agent.name() while InMemoryRunner now defaults to "InMemoryRunner". Updated session creation to use runner.appName(). 

./mvnw -Prelease clean package passes locally.

hemasekhar-p:
@mithun-sudo, thank you for addressing the comments. This PR is currently under review by our team and we will keep you posted if any further information is required.

wikaaaaa:
Hi @mithun-sudo, thank you for the contribution and for addressing the TODO!

Unfortunately, we cannot merge this change. Changing the default `appName` of `new InMemoryRunner(agent)` is a breaking change for existing users.

When callers create a session ahead of time via:
```
Runner runner = new InMemoryRunner(agent);
Session session = runner.sessionService().createSession(agent.name(), userId).blockingGet();
runner.runAsync(userId, session.id(), message);
```
`createSession` writes the session under `agent.name()`, but `runAsync(userId, sessionId, ...)` looks it up under the runner's own `this.appName`. Today those match. If the default flips to "InMemoryRunner", the lookup looks under the new key, misses, and throws `IllegalArgumentException: Session not found` on the first call (as seen in the test updates required by this PR).

`InMemoryRunner` already provides a constructor that accepts an explicit appName:
`new InMemoryRunner(agent, "InMemoryRunner");`

Anyone who specifically needs appName to be "InMemoryRunner" (for example, for cross-language parity) can achieve it that way today without breaking existing callers.

In this case, backwards compatibility for existing Java users is more important than parity with Python. That TODO comment shouldn't have been left in the codebase; we will take care of cleaning it up on our end.

Thank you again for taking the time to send the PR, and sorry for the churn caused by that comment!

PR #1502, opened 2026-09-11
fix(spring-ai): declare ordering for SpringAIAutoConfiguration so @ConditionalOnBean matches Spring AI model beans

### LINKED ISSUE #1501: [google-adk-spring-ai] SpringAI beans are never auto-configured: @ConditionalOnBean(ChatModel) is evaluated before Spring AI model auto-configurations (missing ordering declaration)
# Environment

| Component | Version |
|---|---|
| google-adk / google-adk-spring-ai | 1.9.0 |
| Spring AI (BOM) | 2.0.1 |
| Spring Boot | 4.0.2 |
| Java | 25 |

## Minimal reproduction

**pom.xml** (relevant parts):

```xml
<parent>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-starter-parent</artifactId>
    <version>4.0.2</version>
</parent>

<properties>
    <java.version>25</java.version>
    <spring-ai.version>2.0.1</spring-ai.version>
    <adk.version>1.9.0</adk.version>
</properties>

<dependencies>
    <dependency>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-starter</artifactId>
    </dependency>
    <dependency>
        <groupId>com.google.adk</groupId>
        <artifactId>google-adk</artifactId>
        <version>${adk.version}</version>
    </dependency>
    <dependency>
        <groupId>com.google.adk</groupId>
        <artifactId>google-adk-spring-ai</artifactId>
        <version>${adk.version}</version>
    </dependency>
    <dependency>
        <groupId>org.springframework.ai</groupId>
        <artifactId>spring-ai-starter-model-openai</artifactId>
    </dependency>
</dependencies>

<dependencyManagement>
    <dependencies>
        <dependency>
            <groupId>org.springframework.ai</groupId>
            <artifactId>spring-ai-bom</artifactId>
            <version>${spring-ai.version}</version>
            <type>pom</type>
            <scope>import</scope>
        </dependency>
    </dependencies>
</dependencyManagement>
```

**Application.java:**

```java
@SpringBootApplication
public class DemoApplication {

    public static void main(String[] args) {
        SpringApplication.run(DemoApplication.class, args);
    }

    @Bean
    LlmAgent agent(SpringAI springAI) {
        return LlmAgent.builder()
                .name("demo-agent")
                .model(springAI)
                .instruction("You are a helpful assistant.")
                .build();
    }
}
```

**application.yaml** (dummy values are enough — the failure happens at bean-definition processing, before any remote call):

```yaml
spring:
  ai:
    openai:
      base-url: https://api.openai.com
      api-key: dummy
      chat:
        model: gpt-4o-mini
```

Run: `mvn spring-boot:run`

## Actual behavior

Application fails to start:

```
***************************
APPLICATION FAILED TO START
***************************

Description:

Parameter 0 of method agent in com.example.demo.DemoApplication required a bean
of type 'com.google.adk.models.springai.SpringAI' that could not be found.

Action:

Consider defining a bean of type 'com.google.adk.models.springai.SpringAI' in your configuration.
```

Running with `--debug` shows the root cause — the OpenAI `ChatModel` bean definition **is** registered, but `SpringAIAutoConfiguration` is processed **before** it, so all three `SpringAI` `@Bean` methods are skipped:

```
OpenAiChatAutoConfiguration matched:
   - @ConditionalOnProperty (spring.ai.model.chat=openai) matched
OpenAiChatAutoConfiguration#openAiChatModel matched:
   - @ConditionalOnMissingBean ... did not find any beans (OnBeanCondition)

SpringAIAutoConfiguration#springAIWithBothModels:
   Did not match:
      - @ConditionalOnBean (types: ...ChatModel, ...StreamingChatModel) did not find any beans
SpringAIAutoConfiguration#springAIWithChatModel:
   Did not match:
      - @ConditionalOnBean (types: ...ChatModel) did not find any beans of type ...ChatModel (OnBeanCondition)
SpringAIAutoConfiguration#springAIWithStreamingModel:
   Did not match:
      - @ConditionalOnBean (types: ...StreamingChatModel) did not find any beans
```

Note `springAIEmbedding` has the same problem (`@ConditionalOnBean(EmbeddingModel.class)`) — it fails **silently** (no bean, no error), which makes it even harder to notice.

## Expected behavior
`SpringAIAutoConfiguration` registers a `SpringAI` bean whenever a `ChatModel` (and/or `StreamingChatModel`) is defined by Spring AI model auto-configurations, regardless of which provider starter is used.
## Root cause
`SpringAIAutoConfiguration` guards its `@Bean` methods with `@ConditionalOnBean(...)`, but declares no ordering relative to the auto-configurations that may provide those beans. Per the Spring Boot reference documentation, the result of `@ConditionalOnBean` depends on what has been processed so far, so an auto-configuration relying on it must be ordered after the configurations that register the matched beans.

`google-adk-spring-ai` 1.9.0 is built against Spring AI 2.x / Boot 4 (its POM depends on `spring-ai-model:2.0.1` and `spring-boot-autoconfigure:4.0.2`), and since it compiles only against `spring-ai-model` (no provider modules), the Class-based `after = ...` attribute is not usable here — the string-based `afterName` attribute exists exactly for this case (non-existent class names are silently ignored by the sorter).

## Suggested fix

Add an `afterName` list covering the Spring AI model auto-configurations, mirroring what Spring AI 1.x's own `ChatClientAutoConfiguration` did for the identical scenario (`@ConditionalOnBean(ChatModel)` with an unpredictable provider):

```diff
-@AutoConfiguration
+@AutoConfiguration(afterName = {
+    "org.springframework.ai.model.openai.autoconfigure.OpenAiChatAutoConfiguration",
+    "org.springframework.ai.model.openai.autoconfigure.OpenAiEmbeddingAutoConfiguration",
+    // remaining Spring AI 2.x model auto-configurations, e.g.:
+    // org.springframework.ai.model.anthropic.autoconfigure.AnthropicChatAutoConfiguration
+    // org.springframework.ai.model.ollama.autoconfigure.OllamaChatAutoConfiguration
+    // ... (package pattern: org.springframework.ai.model.<provider>.autoconfigure.*)
+})
 @ConditionalOnClass({SpringAI.class, ChatModel.class})
 @ConditionalOnProperty(prefix = "adk.spring-ai.auto-configuration", name = "enabled",
     havingValue = "true", matchIfMissing = true)
 @EnableConfigurationProperties({SpringAIProperties.class})
 public class SpringAIAutoConfiguration {
```

(Verified against Spring AI 2.0.1: `org.springframework.ai.model.openai.autoconfigure.OpenAiChatAutoConfiguration` / `OpenAiEmbeddingAutoConfiguration`. The other provider names above follow the 2.x package pattern and should be double-checked against the current source.)

An alternative design — the one Spring AI 2.0 itself adopted — is to drop the `@ConditionalOnBean` guards entirely and take `ChatModel`/`StreamingChatModel` as `@Bean` method parameters, which removes the ordering sensitivity altogether. The `afterName` fix is the smaller, backward-compatible change, though.

## Workaround (for users on 1.9.0)

Define the `SpringAI` bean in application code — user-defined beans are processed before auto-configurations, and ADK's `@ConditionalOnMissingBean(SpringAI.class)` then backs off:

```java
@Bean
SpringAI springAI(ChatModel chatModel) {
    return new SpringAI(chatModel, "gpt-4o-mini");
}
```

This works with any provider (it consumes the `ChatModel` abstraction), but every project using ADK + Spring AI has to copy it, so it is not a substitute for the upstream fix.

## References

- Spring Boot reference — [Creating your own auto-configuration / condition annotations](https://docs.spring.io/spring-boot/reference/features/developing-auto-configuration.html): ordering matters for `@ConditionalOnBean`
- Spring AI 1.x `ChatClientAutoConfiguration`: same pattern solved with `afterName` over all provider auto-configurations
- Spring AI 2.0.1 `ChatClientAutoConfiguration`: same problem avoided via `@Bean` parameter injection instead of `@ConditionalOnBean`

kongxubihai:
Fixes #1501

## Problem

`SpringAIAutoConfiguration` (contrib/spring-ai) guards all four of its `@Bean` methods with
`@ConditionalOnBean` — three `SpringAI` variants on `ChatModel`/`StreamingChatModel`, and
`springAIEmbedding` on `EmbeddingModel` — but declares no ordering relative to the
auto-configurations that register those beans.

The result of `@ConditionalOnBean` depends on what has been processed so far. With no
declared ordering, evaluation order falls back to the alphabetical class-name sort, and
`com.google.adk...` sorts before `org.springframework.ai...` — so ADK's conditions are
evaluated before any Spring AI model bean definition exists. No `SpringAI` bean is
registered and applications fail to start:

```
Parameter 0 of method scienceTeacher in ...AdkJavaSpringAiDemoApplication required a bean
of type 'com.google.adk.models.springai.SpringAI' that could not be found.
```

`springAIEmbedding` is affected identically — silently: no error, just a missing bean.

## Fix

Declare `afterName` on `@AutoConfiguration`, listing the Spring AI model auto-configurations
that may provide `ChatModel`, `StreamingChatModel`, or `EmbeddingModel` beans
(7 chat + 9 embedding entries in the commit).

Design notes:

1. **Why `afterName` (string) and not `after` (Class):** this module compiles only against
   `spring-ai-model`; provider auto-configuration classes are not on the compile classpath.
   Names that do not exist on the classpath are ignored by the sorter, which makes string
   form the intended mechanism here — the same pattern Spring AI 1.x's own
   `ChatClientAutoConfiguration` used for the identical problem.
2. **How the list was derived and verified** (against the published Spring AI 2.0.1 jars):
   - Read every provider module's
     `META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports`
     (the authoritative registration file — class-path/name-pattern guessing misses
     nested packages such as `google.genai.autoconfigure.chat` or `vertexai.autoconfigure.embedding`).
   - `javap`-verified the bean type of every configuration whose name alone is not
     sufficient: `GoogleGenAiChatModel implements ChatModel`;
     `GoogleGenAiTextEmbeddingModel` / `VertexAiTextEmbeddingModel` extend
     `AbstractEmbeddingModel`.
3. **Deliberately excluded** (each verified, not guessed):
   - `*Connection*AutoConfiguration` (google-genai, vertex-ai) — register connection-details
     beans, not models;
   - `VertexAiMultiModalEmbeddingAutoConfiguration` — registers a `DocumentEmbeddingModel`,
     which does not satisfy `@ConditionalOnBean(EmbeddingModel)`;
   - `ElevenLabsAutoConfiguration` — registers a text-to-speech model, not chat/embedding;
   - `OllamaApiAutoConfiguration` and the image/audio/moderation/OCR configurations —
     unrelated interfaces.
4. **Maintenance:** new Spring AI provider modules need to be appended to this list — the
   same standing cost Spring AI 1.x accepted for this pattern. Modules introduced after
   2.0.1 are ignored harmlessly until added.

## Test

New `SpringAIAutoConfigurationOrderingTest` places `SpringAIAutoConfiguration` first in
`AutoConfigurations.of(...)` alongside the provider configurations and
`ToolCallingAutoConfiguration` (the latter provides the `ToolCallingManager` that
`OpenAiChatAutoConfiguration` requires; a real application imports it automatically, the
context runner must declare it explicitly).

`AutoConfigurations.of` applies the same ordering rules as production auto-configuration
import (`AutoConfigurationSorter#getInPriorityOrder`), so:

- **Before this change** the tests fail: the alphabetical order processes
  `SpringAIAutoConfiguration` before the provider configurations, so `@ConditionalOnBean`
  never matches.
- **After** both tests pass, and runtime behavior flips accordingly — in a demo application,
  singleton pre-instantiation logs change from `springAIEmbedding` being created before
  `openAiChatModel` (alphabetical order, conditions evaluated too early) to OpenAI beans
  being created first followed by `Auto-configuring SpringAI...`.

## Notes

An alternative fix — dropping `@ConditionalOnBean` in favor of `@Bean` method parameter
injection, as Spring AI 2.0's own `ChatClientAutoConfiguration` does — would also remove
the ordering sensitivity, but changes semantics: with no model present, failure moves from
a clear startup report to a less friendly bean-creation error. The `afterName` declaration
is the smaller, behavior-preserving change.

Google CLA signed.

### DIFF
--- a/contrib/spring-ai/pom.xml
+++ b/contrib/spring-ai/pom.xml
@@ -137,6 +137,11 @@
             <artifactId>spring-ai-ollama</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.ai</groupId>
+            <artifactId>spring-ai-starter-model-openai</artifactId>
+            <scope>test</scope>
+        </dependency>
 
         <!-- Testcontainers for local testing -->
         <dependency>
--- a/contrib/spring-ai/src/main/java/com/google/adk/models/springai/autoconfigure/SpringAIAutoConfiguration.java
+++ b/contrib/spring-ai/src/main/java/com/google/adk/models/springai/autoconfigure/SpringAIAutoConfiguration.java
@@ -58,7 +58,28 @@ import org.springframework.context.annotation.Primary;
  * adk.spring-ai.validation.enabled=true
  * </pre>
  */
-@AutoConfiguration
+@AutoConfiguration(
+    afterName = {
+      // --- Spring AI 2.0.1 chat model auto-configurations (verified against published jars) ---
+      "org.springframework.ai.model.anthropic.autoconfigure.AnthropicChatAutoConfiguration",
+      "org.springframework.ai.model.bedrock.converse.autoconfigure.BedrockConverseProxyChatAutoConfiguration",
+      "org.springframework.ai.model.deepseek.autoconfigure.DeepSeekChatAutoConfiguration",
+      "org.springframework.ai.model.google.genai.autoconfigure.chat.GoogleGenAiChatAutoConfiguration",
+      "org.springframework.ai.model.mistralai.autoconfigure.MistralAiChatAutoConfiguration",
+      "org.springframework.ai.model.ollama.autoconfigure.OllamaChatAutoConfiguration",
+      "org.springframework.ai.model.openai.autoconfigure.OpenAiChatAutoConfiguration",
+      // --- embedding model auto-configurations (required for springAIEmbedding,
+      //     which silently fails to register without these) ---
+      "org.springframework.ai.model.bedrock.cohere.autoconfigure.BedrockCohereEmbeddingAutoConfiguration",
+      "org.springframework.ai.model.bedrock.titan.autoconfigure.BedrockTitanEmbeddingAutoConfiguration",
+      "org.springframework.ai.model.google.genai.autoconfigure.embedding.GoogleGenAiTextEmbeddingAutoConfiguration",
+      "org.springframework.ai.model.mistralai.autoconfigure.MistralAiEmbeddingAutoConfiguration",
+      "org.springframework.ai.model.ollama.autoconfigure.OllamaEmbeddingAutoConfiguration",
+      "org.springframework.ai.model.openai.autoconfigure.OpenAiEmbeddingAutoConfiguration",
+      "org.springframework.ai.model.postgresml.autoconfigure.PostgresMlEmbeddingAutoConfiguration",
+      "org.springframework.ai.model.transformers.autoconfigure.TransformersEmbeddingModelAutoConfiguration",
+      "org.springframework.ai.model.vertexai.autoconfigure.embedding.VertexAiTextEmbeddingAutoConfiguration"
+    })
 @ConditionalOnClass({SpringAI.class, ChatModel.class})
 @ConditionalOnProperty(
     prefix = "adk.spring-ai.auto-configuration",
--- a/contrib/spring-ai/src/test/java/com/google/adk/models/springai/autoconfigure/SpringAIAutoConfigurationOrderingTest.java
+++ b/contrib/spring-ai/src/test/java/com/google/adk/models/springai/autoconfigure/SpringAIAutoConfigurationOrderingTest.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.adk.models.springai.autoconfigure;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.ai.model.openai.autoconfigure.OpenAiChatAutoConfiguration;
+import org.springframework.ai.model.openai.autoconfigure.OpenAiEmbeddingAutoConfiguration;
+import org.springframework.ai.model.tool.autoconfigure.ToolCallingAutoConfiguration;
+import org.springframework.boot.autoconfigure.AutoConfigurations;
+import org.springframework.boot.test.context.runner.ApplicationContextRunner;
+
+/**
+ * Regression tests: SpringAIAutoConfiguration guards its beans with {@code @ConditionalOnBean},
+ * which is order-sensitive. It must therefore declare itself ordered after the Spring AI model
+ * auto-configurations, otherwise the conditions never match and no SpringAI/SpringAIEmbedding bean
+ * is registered.
+ */
+class SpringAIAutoConfigurationOrderingTest {
+
+  // OpenAI auto-configurations fail at context refresh without a non-blank api key.
+  private static final String[] PROPERTIES = {"spring.ai.openai.api-key=dummy"};
+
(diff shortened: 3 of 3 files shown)


hemasekhar-p:
Hi @kongxubihai, thank you for your contribution. We appreciate you taking the time to submit this pull request. Currently this PR is under review by our team and we will keep you posted if any additional information is required. thank you.

PR #1515, opened 2026-09-16
fix(mcp): refuse MCP tools that shadow reserved framework names

### LINKED ISSUE #1513: MCP toolset: in-model built-ins (e.g. google_search) can be shadowed by a server tool; a server tool named set_model_response aborts the run
## Summary

Same two defects as the Go port, reproduced locally against `adk-java` (`4ee155b2`, core). No live service was contacted.

1. **In-model built-in tools do not occupy their name, so a server-provided tool can shadow them.** `GoogleSearchTool` (and `GoogleMapsTool`, `UrlContextTool`, `VertexAiSearchTool`, `BuiltInCodeExecutionTool`) override `processLlmRequest` to append only to `config.Tools`, never calling `appendTools`. So the duplicate-name guard in `LlmRequest.Builder.appendTools` never sees them, and a callable MCP tool advertising `google_search` is accepted and wins dispatch — `Functions.handleFunctionCalls` resolves by name from a map where the MCP tool overwrote the built-in.
2. **A server-provided tool named `set_model_response` aborts the whole run** on an agent with an output schema (`IllegalArgumentException: Duplicate tool name: set_model_response`). Availability, not silent displacement — but a malicious/untrusted MCP server can deny service.

## Evidence

- MCP tools take the server's name, unfiltered by default — `tools/mcp/McpToolset.java:260-279`; `tools/BaseToolset.java:60-64` (`if (toolFilter == null) return true;`); name source `tools/mcp/AbstractMcpTool.java:52-56`.
- Callable-vs-callable is fail-closed — `models/LlmRequest.java:204-222` (`appendTools`, throwing merger); all tools enter via `tools/BaseTool.java:187-193`.
- `set_model_response` is in the guarded class — `flows/llmflows/OutputSchema.java:60-65`; `SetModelResponseTool.java:33-45` (extends `BaseTool`, does not override `processLlmRequest`). Processor order: `SingleFlow.java:27-36`, `BaseLlmFlow.java:103-104`.
- Built-ins bypass the guard — `tools/GoogleSearchTool.java:45-76` (override that appends `Tool.builder().googleSearch(...)` only); same in `GoogleMapsTool.java:65`, `UrlContextTool.java:46`, `VertexAiSearchTool.java:60`, `BuiltInCodeExecutionTool.java:44`.
- Dispatch by name — `BaseLlmFlow.java:765-766`; `Functions.java:295` (`tools.get(functionCall.name().get())`).

## Local reproduction

```
[probe] registered tool map keys: [google_search]
[probe]   config.Tool googleSearch=true functionDeclarations=[]
[probe]   config.Tool googleSearch=false functionDeclarations=[google_search]
[probe] google_search: NO ERROR. dispatch map maps google_search -> FakeMcpTool
[probe] set_model_response: REJECTED -> Duplicate tool name: set_model_response
```

## Suggested fix

Preferred (availability-first): reject the server-provided tool via a reserved-name check in `AbstractMcpTool`'s constructor or in `McpToolset.getTools` / `McpAsyncToolset.getTools`, covering `set_model_response`, `transfer_to_agent`, `google_search`, `google_maps`, `url_context`, `vertex_ai_search`, `code_execution`, `load_artifacts`.

Alternative: make the in-model tools call `appendTools` so the existing duplicate guard applies (note: `google_search` would then need a clean-failing `runAsync`).

Same class as the Go port (`google/adk-go#1605`) and a separately-triaged `adk-js` report.

sushant-me:
Fixes #1513

## Problem

`McpToolset` accepted any tool name an MCP server advertised. Names that the
ADK framework reserves for its own built-ins were therefore installable by a
remote server, and the duplicate-name guard in
`LlmRequest.Builder.appendTools` never noticed:

* In-model built-ins (`google_search`, `google_maps`, `url_context`,
  `vertex_ai_search`, `code_execution`) are appended only to the request's
  `config.tools` and are never placed in the tool map, so they never collide
  with the MCP entry.
* `set_model_response` is consumed internally by
  `SingleTurnAgentExecutor`/`LlmFlow`; a server tool of that name makes the
  flow abort the whole run where it expects its own response model.

The result is that an MCP server silently wins dispatch for a framework tool
name — a tool-boundary bug, not a naming nit.

## Fix

Reserve the names in `McpToolset.getTools()` and fail loading with
`McpToolLoadingException` when a server advertises one:

```java
private static final Set<String> RESERVED_TOOL_NAMES = Set.of(
    "set_model_response", "transfer_to_agent", "google_search", "google_maps",
    "url_context", "vertex_ai_search", "code_execution", "load_artifacts",
    "loadMemory", "exit_loop", "list_skills", "load_skill",
    "load_skill_resource");
```

**Correction (later push).** An earlier revision of this PR listed the Go set
verbatim, which was wrong in both directions.

* `finish_task` and `task_completed` do not exist anywhere in this port. They
  were refused as collisions against tools this framework does not have.
* The memory tool is named `loadMemory` here, not `load_memory`. This port takes
  a tool's name from the method name when the method carries no
  `@Annotations.Schema` — `LoadMemoryTool#loadMemory` annotates only its
  parameter — so the other ports' `load_memory` spelling did not match, and the
  guard was not watching for the name it actually uses.

Two tests now pin this: `getTools_refusesDerivedLoadMemoryName`, and
`getTools_acceptsNamesOtherPortsDefineButThisOneDoesNot`, which asserts that the
three names from the other ports are **accepted** rather than refused.

`McpToolLoadingException` is used deliberately: `retryWhen` treats
`IllegalArgumentException` as fatal, and a config error of this kind should
surface once rather than be retried as a transient server fault.

## Tests

`getTools_refusesReservedToolName` mocks an `McpSyncClient` returning
`McpSchema.Tool.builder().name("google_search")` and asserts the load fails
with `McpToolLoadingException`. It also verifies the client is called exactly
once, pinning the no-retry behaviour.

Verified in both directions:

* without the change to `McpToolset.java`: `Tests run: 1, Failures: 1`
* with the change: green

This mirrors the equivalent `google/adk-go` fix in #1606.

## Why a list at all, and why I would rather not keep one

This is the second correction to `RESERVED_TOOL_NAMES`, and the reason is the
approach rather than either edit. The set was assembled from names reported one
at a time, so each fix addressed the name in front of it and left the rest. Only
after enumerating the tools this framework actually declares — rather than
transcribing a list — did the remaining four appear.

The structural alternative is better and I would rather implement it: give the
in-model tools their name in the request's tool map, so the existing duplicate-name
check in `LlmRequest.appendTools` refuses a server that takes one. That guard
already exists and already has the right semantics; the in-model tools currently
bypass it by appending only to the config tools. A list needs maintaining on every
tool the framework adds, and its completeness is not checkable by the same
mechanism that uses it.

If you would rather have the list, it is here and tested. If you would rather have
the guard apply at the point where names are already being checked, I am happy to
rework it that way instead.


hemasekhar-p:
Hi @sushant-me , thank you for your contribution. We appreciate you taking the time to submit this pull request. Currently this PR is under review by our team, we will keep you posted if any additional information is required. thank you.

sushant-me:
One scoping note on the residual, so the trade-off is on the record — the same class of note as the one on the Go port, but the reasoning differs here and I did not want to copy it across.

This guard closes the path where a **remote MCP server** advertises a reserved name. It does not change the underlying property that an in-model tool never occupies its name in the tool map:

* `GoogleSearchTool.processLlmRequest` (and the other in-model tools) only append to `configBuilder.tools(...)`; they never put the name into `llmRequestBuilder.tools()`.
* Nothing else does either — I could not find a duplicate-name check at the `tools()` map level in the flows (the only duplicate check is for sub-agent names in `BaseAgent`).

So an in-process tool registered under `google_search` — a `FunctionTool` or an `AgentTool` — is accepted alongside the in-model tool. In this repository that requires the application author to pick the name, so I treated it as a footgun rather than the adversarial case and kept this PR to the server-controlled path.

**Where Java differs from Go:** in `adk-go` the equivalent map is `map[string]any` and consumers type-assert to `tool.Tool`, so occupying a name with a sentinel breaks them — that is why I left the general fix alone there. Here `LlmRequest.tools()` is already `Map<String, BaseTool>` (LlmRequest.java:89), so registering a sentinel or a real holder under the in-model name is type-safe. The open question is only whether a sentinel value is acceptable to advertise, which is a design decision rather than a typing constraint.

Happy to prepare that broader change if you would prefer the invariant enforced in one place instead of at each boundary — it touches high-fan-in classes, so I did not want to fold it in unasked.

sushant-me:
Thanks, @hemasekhar-p — appreciated.

One thing that may help while it is in review: `RESERVED_TOOL_NAMES` currently carries **all 13** names Java puts on the wire, and I checked it is neither short nor over-broad — every entry is a `super("...")` literal in a non-test source, plus `set_model_response` and `transfer_to_agent` which the framework contributes outside a tool class. Two names an earlier revision of my notes carried over from the Go port (`finish_task`, `task_completed`) are **not** Java names and are deliberately absent.

The two halves of #1513 fail differently in this code, which is why the guard covers both:

- **In-model built-ins** (`google_search`, `google_maps`, `url_context`, `vertex_ai_search`, `code_execution`) append only to `config.Tools` and never reach `appendTools`, so a callable tool takes the name and `Functions.handleFunctionCalls` resolves it to the server tool.
- **Callables** (`set_model_response` and friends) are already fail-closed — `appendTools` throws `Duplicate tool name` — so there the guard converts a run-aborting exception into a clean registration error.

Happy to adjust the list, the error type, or the placement if your team prefers a different shape.

